Issue #17, opened 2019-02-17
Sitreps and Chosen panels visible on mission infiltration screens


pledbrook:
See #16 for a screenshot demonstrating the problem. The screen is displaying panels that don't apply to the mission, i.e. active Chosen and sitreps. I tried to fix the sitreps panel [here](https://github.com/pledbrook/lwotc/blob/master/LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2StrategyElement_LWMissionSources.uc#L73), but it doesn't seem to be working.

tracktwo:
This one should be resolved now too, I think.

pledbrook:
OK, cool. I just want to check before closing.

PR #20, opened 2019-02-19
Fix several issues with the GOp mission UI

### LINKED ISSUE #16: Mission infiltration and launch screens show confirm button in wrong place
screenshot 1

See how the Confirm button appears at the top of the screen on the right.

tracktwo:
Fixes #16 and #17

The interaction between the base `UIMission` and mission-specific
subtypes changed in WotC and the LW version needed some updates. This
change goes through the `UIMission_GOps` class from WOTC and updates the
LW version to be roughly equivalent.

In particular the class needs an empty `OnButtonSizeRealized` function
to avoid the base version moving the confirm button around (#16), and
the screen building needs to invoke the base class `BuildScreen` to
properly set up the chosen and sitrep UI elements (among others) so
they will be hidden when appropriate (#17).

It's probably better to split this file up into separate versions for
different mission types, but I didn't do that here. Non-GOp missions
may still have bugs because of this. Dark Event display may need
updating too.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIMission_LWCustomMission.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIMission_LWCustomMission.uc
@@ -5,6 +5,11 @@
 //			 This is used for initiating infiltration/investigation of a mission site 
 //			 Launching a mission after investigation has begun is handled in UIMission_LWLaunchDelayedMission
 //--------------------------------------------------------------------------------------- 
+
+// WOTC TODO - This file is probably better split up into separate versions for particular mission sub-types like
+// they are in WOTC. These sub-classes involve many override functions called from the base UIMission and may need
+// to take different actions depending on which mission type is involved.
+
 class UIMission_LWCustomMission extends UIMission config(LW_Overhaul);
 
 enum EMissionUIType
@@ -175,21 +180,26 @@ simulated function String GetMissionTitle()
 
 simulated function BuildScreen()
 {
-	// Add Interception warning and Shadow Chamber info 
-	BindLibraryItem();
-	`LWACTIVITYMGR.UpdateMissionData(GetMission());
-	BuildMissionPanel();
-	BuildOptionsPanel();
-	class'UIUtilities_LW'.static.BuildMissionInfoPanel(self, MissionRef);
-
 	PlaySFX(GetSFX());
-
 	XComHQPresentationLayer(Movie.Pres).CAMSaveCurrentLocation();
+	`HQPres.StrategyMap2D.HideCursor();
 
 	if(bInstantInterp)
 		XComHQPresentationLayer(Movie.Pres).CAMLookAtEarth(GetMission().Get2DLocation(), CAMERA_ZOOM, 0);
 	else
 		XComHQPresentationLayer(Movie.Pres).CAMLookAtEarth(GetMission().Get2DLocation(), CAMERA_ZOOM);
+
+	// Add Interception warning and Shadow Chamber info
+	`LWACTIVITYMGR.UpdateMissionData(GetMission());
+
+	// Base version is responsible for showing most mission info, including the mission and options panels,
+	// shadow chamber, chosen, sitreps, etc.
+	super.BuildScreen();
+
+	// Add the LW-specific mission info: squad size, concealment type, evac mode, etc.
+	class'UIUtilities_LW'.static.BuildMissionInfoPanel(self, MissionRef);
+
+	BuildConfirmPanel();
 }
 
 // Called when screen is removed from Stack
@@ -316,6 +326,28 @@ simulated function AddIgnoreButton()
 	{
 		IgnoreButton.InitButton('IgnoreButton').Hide();
 	}
+}
+
+simulated function OnButtonSizeRealized()
+{
+	// Override - do nothing
+}
+
+simulated function UpdateData()
+{
+	// WOTC TODO different sfx for different missions?
+	`XSTRATEGYSOUNDMGR.PlaySoundEvent("Geoscape_AlienOperation");
+	XComHQPresentationLayer(Movie.Pres).CAMLookAtEarth(GetMission().Get2DLocation(), CAMERA_ZOOM);
+
+	BuildMissionPanel();
+	//RefreshNavigation();
+
+	super.UpdateData();
+}
+
+simulated function bool CanBackOut()
+{
+	return (super.CanBackOut() && class'XComGameState_HeadquartersXCom'.static.IsObjectiveCompleted('T0_M7_WelcomeToGeoscape'));
 }
 
 // ----------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIMission_LWCustomMission.uc
```diff
@@ -5,6 +5,11 @@
 //			 This is used for initiating infiltration/investigation of a mission site 
 //			 Launching a mission after investigation has begun is handled in UIMission_LWLaunchDelayedMission
 //--------------------------------------------------------------------------------------- 
+
+// WOTC TODO - This file is probably better split up into separate versions for particular mission sub-types like
+// they are in WOTC. These sub-classes involve many override functions called from the base UIMission and may need
+// to take different actions depending on which mission type is involved.
+
```
pledbrook: Agreed and I'll raise an issue for that.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIMission_LWCustomMission.uc
```diff
@@ -318,6 +328,28 @@ simulated function AddIgnoreButton()
 	}
 }
 
+simulated function OnButtonSizeRealized()
```
pledbrook: In my version of this fix I added a comment that explained why this was empty. I can add it in later if you'd like. I like to explain unusual code so that others are aware of gotchas and don't try to prune stuff like this.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIMission_LWCustomMission.uc
```diff
@@ -501,15 +537,17 @@ simulated function BuildGuerrillaOpsOptionsPanel()
...
 	LibraryPanel.MC.QueueString(""); //GetGOpsMissionLocString(2));
 	LibraryPanel.MC.QueueString(class'UIUtilities_Text'.default.m_strGenericConfirm);
-	LibraryPanel.MC.QueueString(bCanBackOut ? m_strIgnore : ""); // defined in UIX2SimpleScreen
+	LibraryPanel.MC.QueueString(CanBackOut() ? m_strIgnore : ""); // defined in UIX2SimpleScreen
 	LibraryPanel.MC.EndOp();
 
 	// ----------------------------------------------------------------------
 
+	// WOTC TODO Other mission types may be missing these
 	BuildConfirmPanel();
```
pledbrook: What do you think of removing this line? It doesn't do anything unless `LibraryPanel` is `none`, which should never be the case.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIMission_LWCustomMission.uc
```diff
@@ -175,21 +180,26 @@ simulated function String GetMissionTitle()
...
+	`LWACTIVITYMGR.UpdateMissionData(GetMission());
+
+	// Base version is responsible for showing most mission info, including the mission and options panels,
+	// shadow chamber, chosen, sitreps, etc.
+	super.BuildScreen();
+
+	// Add the LW-specific mission info: squad size, concealment type, evac mode, etc.
+	class'UIUtilities_LW'.static.BuildMissionInfoPanel(self, MissionRef);
+
+	BuildConfirmPanel();
```
pledbrook: Is this worth it? I'm pretty sure it doesn't do anything. I think it's confusing to call a method that looks like it should do something with the Confirm and Ignore buttons, but in fact does nothing. Have I missed something?

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIMission_LWCustomMission.uc
tracktwo: It is apparently useless, yeah. The only reason I kept it was for symmetry with the original version. That's probably a lot less of an issue now unless they make another xpack or more DLC, but when that happens it's helpful to see exactly what changed from the base game versions and why. I can also just comment it out, with a comment explaining why.

tracktwo:
> Thanks for this!
> 
> Would you mind including `UIMission_LWLaunchDelayedMission`? It basically needs the same set of changes. It does have an additional problem of the "time left" information overlapping the Launch Mission button, but I'm happy to tackle that as a separate issue. If you don't have time, no worries.

Can do. I totally forgot there was another copy. I also suspect more changes will be needed here. It's been a long time since I played wotc but I think the mission info panel placement will need to change when a chosen is active. I will try to test it out and see what it looks like. There's a whole lot going on in these screens when dark events, sitreps, and chosen are active. I'm not sure where there will be space to put the mission info stuff.

pledbrook:
I'm thinking that a new panel will be needed, perhaps with a transparent background. Centre top. Not sure. But yes, definitely short of space in the main mission panels.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIMission_LWCustomMission.uc
pledbrook: Sounds good to me.

tracktwo:
I've got the launch mission screen mostly working, but there is another hitch. The original XCOM2 buttons panel had blank space between the mission options buttons and the confirm/ignore buttons. LW2 used this empty space to show the infiltration status text (% infiltrated, time to 100%, enemy strength). WOTC shrunk this panel to make more room for the sitrep panel, eliminating this dead space, so now the infiltration text overlaps the confirm/ignore buttons. Will need to find a new place to put this text.

Attached is a screenshot that shows both the bad infil text and the mission brief overlapping the chosen info. Can probably solve both by moving the mission brief up a bit and adding the infiltration text to it. The mission brief was added later in LW2, if we had it from the start it probably would've been the logical place to put the infil status in the first place. The time to expiry is repeated in both places,for example.
bad_infil_text

pledbrook:
I'd be tempted to move Mission Brief to above the options panel and include the % infiltration and enemy readiness as you suggest. Happy to trust your judgement on this.

tracktwo:
The shadow chamber info will go above the options panel, so it can't go there. It can almost fit above the chosen without clipping the green reticle in the centre but it does overlap the pulsing red Chosen foggy stuff. I moved it top centre instead. It still seems too busy to me, and it's going to be crazy with both shadow chamber and sitrep stuff on there.

This is the kind of thing that can be tweaked endlessly, so I just have one more small change to make to the spacing when you aren't currently infiltrating and call it good enough for now. There is still some duplicated info here, the "Current enemy activity" and "Enemy readiness" are two different pieces of related information. That was a minor issue I had with the system before, and one a lot of people had when they weren't familiar with the infiltrating mechanics and were wondering why the mission said "extremely light" but then all of a sudden said "heavy".

better_infil

tracktwo:
It also makes my eye twitch a little bit that the infil timer says "Xd Yh" until 100% but the expiry says "X Days, Y Hours". :)

pledbrook:
I pretty much concur with everything you've said, including the eye twitching :smile: I'm sure we can improve this, but it's not a priority.

pledbrook:
Thanks for this BTW.

PR #41, opened 2019-03-28
Fix Red Fog

tracktwo:
`GetHPChangedObjectList` was pulling the previous game state with an
uninitialized variable, so would always yield `none` and ignore all
HP-changing events.


pledbrook:
Just saw this. Thanks!

PR #143, opened 2019-04-29
Modified XComLWOverhaul.ini for some weapons

jmartinez989:
Made Spark rifles have 3 slots for weapon upgrades as they do int WotC.
Also made changes to required techs for mag versions of Templar and
Skirmisher weapons. Below are their new requirements:

    Templar Gaunlets:
        Plated Armor
        Advent Stun Lancer Autopsy

    Skirmisher Ripjack:
        Advent Stunlancer Autopsy

Made these changes to be consistent with other melee weapon upgrade
paths. Also made sure to make all of the faction unit beam weapons cost
1 elerium core each.


pledbrook:
Should T3 gauntlets and ripjacks also have their requirements updated?

jmartinez989:
T3 Gauntlets and Ripjacks were already ok. Gauntlets had requirements of archon autopsy and powered armor where ripjack just had archon autopsy. Also the gauntlet had the 1 core cost but the ripjack did not so I updated that.

pledbrook:
I think requiring an elerium core for the T2 weapons is over the top. I don't think any other weapons of that tier require it. Certainly not the sword.

jmartinez989:
no I didnt make any of the t2 faction weapons cost a core, only the beam ones. Made sure to double check that in the ini. As far as I can see none of the t2 faction weapons have a core cost.

LongWarOfTheChosen/Config/XComLW_Overhaul.ini
```diff
@@ -597,15 +597,15 @@ INSTANT_BUILD_TIMES=TRUE
...
-+ItemTable=(ItemTemplateName="WristBlade_MG",	Slots=0,	Starting=false,	Infinite=false,	Buildable=true,	RequiredTech1="MagnetizedWeapons",		RequiredTech2="",	SupplyCost=30,	AlloyCost=3,	CrystalCost=3,	CoreCost=0,	SpecialItemTemplateName="CorpseAdventStunLancer",	SpecialItemCost=1,	TradingPostValue=15, RequiredEngineeringScore=10, Tier = 150, PointsToComplete=100, InventoryImage="img:///UILibrary_XPACK_StrategyImages.Inv_MagSGauntlet")
++ItemTable=(ItemTemplateName="ShardGauntlet_MG",	Slots=0,	Starting=false,	Infinite=false,	Buildable=true,	RequiredTech1="PlatedArmor",		RequiredTech2="AutopsyAdventStunLancer",	SupplyCost=35,	AlloyCost=3,	CrystalCost=3,	CoreCost=0,	SpecialItemTemplateName="CorpseAdventStunLancer",	SpecialItemCost=1,	TradingPostValue=15, RequiredEngineeringScore=10, Tier = 150, PointsToComplete=100, InventoryImage="img:///UILibrary_XPACK_StrategyImages.Inv_MagTGauntlet")
```
pledbrook: Why tie the Shard Gauntlet to armour tech? It's a bit confusing considering none of the other weapon techs have such a requirement.

LongWarOfTheChosen/Config/XComLW_Overhaul.ini
```diff
@@ -597,15 +597,15 @@ INSTANT_BUILD_TIMES=TRUE
...
-+ItemTable=(ItemTemplateName="WristBlade_BM",	Slots=0,	Starting=false,	Infinite=false,	Buildable=true,	RequiredTech1="AutopsyArchon",		RequiredTech2="",	SupplyCost=50,	AlloyCost=4,	CrystalCost=4,	CoreCost=0,	SpecialItemTemplateName="CorpseArchon",	SpecialItemCost=1,	TradingPostValue=25, RequiredEngineeringScore=10, Tier = 150, PointsToComplete=100, InventoryImage="img:///UILibrary_XPACK_StrategyImages.Inv_BeamSGauntlet")
++ItemTable=(ItemTemplateName="ShardGauntlet_BM",	Slots=0,	Starting=false,	Infinite=false,	Buildable=true,	RequiredTech1="AutopsyArchon",		RequiredTech2="PoweredArmor",	SupplyCost=55,	AlloyCost=4,	CrystalCost=4,	CoreCost=1,	SpecialItemTemplateName="CorpseArchon",	SpecialItemCost=1,	TradingPostValue=25, RequiredEngineeringScore=10, Tier = 150, PointsToComplete=100, InventoryImage="img:///UILibrary_XPACK_StrategyImages.Inv_BeamTGauntlet")
```
pledbrook: Same as above.

jmartinez989:
Well the reason I went with the armor requirements is because the armor is required for them in vanilla wotc as well. You get temptest gauntlets after plated armor and celestial gauntlets after powered armor.

I kind of makes sense that it is needed since the gauntlets are technically part of their armor. The corpse and autopsy requirements I'll leave to you. I def think that at least the m3 ones should still cost a core.

pledbrook:
The thing is, you can equip T2 Shard Gauntlet with T1 armour, so it just seems weird that the former requires plated armour as a tech. I don't really understand why vanilla made that connection. If the gauntlet were _part_ of the armour, I could understand. But it's not.

I'd prefer to deviate from vanilla here for consistency with all other weapons in the game.

jmartinez989:
Okie doke, Will remove the armor requirement, will just keep the autopsy and corpse costs.

pledbrook:
Another thing to consider is that it makes those two faction soldiers weaker than Reapers (as if they needed any more relative buffs!), since Reapers would be able to get their weapon upgrade sooner. And at least Skirmishers get their Bullpup with the Mag-tier weapons.

jmartinez989:
It's for that reason that I gave the gauntlets the biggest power spike at t3. And with the way the tech tree is right now, it is more likely you will unlock t3 gauntlets before other weapons.

LongWarOfTheChosen/Config/XComLW_Overhaul.ini
jmartinez989: Removed armor tech requirement.

LongWarOfTheChosen/Config/XComLW_Overhaul.ini
jmartinez989: Removed armor tech requirement.

jmartinez989:
Okay, T2 and T3 templar gauntlets no longer require the armor.

LongWarOfTheChosen/Config/XComLW_Overhaul.ini
```diff
@@ -597,14 +597,14 @@ INSTANT_BUILD_TIMES=TRUE
...
 +ItemTable=(ItemTemplateName="VektorRifle_MG",	Slots=3,	Starting=false,	Infinite=false,	Buildable=true,	RequiredTech1="MagnetizedWeapons",		RequiredTech2="",	SupplyCost=50,	AlloyCost=3,	CrystalCost=3,	CoreCost=0,	SpecialItemTemplateName="",	SpecialItemCost=0,	TradingPostValue=25, RequiredEngineeringScore=10, Tier = 150, PointsToComplete=100, InventoryImage="img:///UILibrary_XPACK_Common.MagVektor_Base")
-+ItemTable=(ItemTemplateName="ShardGauntlet_MG",	Slots=0,	Starting=false,	Infinite=false,	Buildable=true,	RequiredTech1="PlatedArmor",		RequiredTech2="AutopsyAdventStunLancer",	SupplyCost=35,	AlloyCost=3,	CrystalCost=3,	CoreCost=0,	SpecialItemTemplateName="CorpseAdventStunLancer",	SpecialItemCost=1,	TradingPostValue=15, RequiredEngineeringScore=10, Tier = 150, PointsToComplete=100, InventoryImage="img:///UILibrary_XPACK_StrategyImages.Inv_MagTGauntlet")
```
pledbrook: Is it OK to have no Tech1 requirement but have a Tech2 requirement? I genuinely don't know, so want to make sure it's OK.

pledbrook:
> It's for that reason that I gave the gauntlets the biggest power spike at t3. And with the way the tech tree is right now, it is more likely you will unlock t3 gauntlets before other weapons.

Why will they unlock before other weapons? Because it doesn't have a requirement on plasma weapons? I'm not sure it matters that much. Aren't the Celestial Gauntlets fairly weak compared to a Shinobi with a fusion blade for example? In fact, aren't all the later faction weapons fairly weak for LW2?

An alternative is to make plasma a requirement for the Celestial Gauntlets. But you'd then have to make Mag weapons a prerequisite for Tempest Gauntlets. I'm not too bothered either way at this point. I just don't want players banging their heads against a brick wall because they can't work out why they can't upgrade Shard Gauntlets while other weapons can be.

jmartinez989:
Well if we lock celestial behind the same tech as all of the other melee weapons then it would be after completing archon autopsy. It is possible to research plasma before that but the resource requirements for plasma are so high (because you need so many cores) that it is likely you'll have Archon autopsy researched first.

Also I boosted the  damage of all of the faction weapons by quite a bit at the late game. The Templar T3 gauntlets now do 10-11 base where they did 6-10 before. That is a bump of +4 on the base damage. That is without focus. With full focus at max rank they would be doing 13-14 which is higher than most other beam weapons. Figured since they have to put themselves at risk the most to do any damage they should get the highest bump.

I kept the bullpups at the weakest strength boost wise because they also benefit from lower detection range and they can shoot twice. I am thinking about bumping up beam ripjack damage up 9 or maybe even 10 to match other melee weapons more closely.

pledbrook:
Wow, that is a big buff. They'll be one-shotting their favoured Chosen.

jmartinez989:
well that will bring us to the topic of balancing the chosen out more to have more health at higher tiers though that would be a topic for when we implement the chosen :P

LongWarOfTheChosen/Config/XComLW_Overhaul.ini
jmartinez989: T1 do not have tech requirements as they are available from the start with no research. Also removed the armor requirements for t2.

jmartinez989:
Actually before you merge, I am going to increase the ripjack dmage. I think 7 for t3 is a bit low. Going to increase it to 10.

Also speaking of balance, chosen weapons are going to be kind of a problem, particularly the Darklance because they will give all snipers with death from above free serial and snap shot as it only costs 1 action point to fire that. Again that is a topic for another day, just making sure we do realize that the balance needs to be done.

Issue #139, opened 2019-04-29
Cannot infiltrate alien facility


jmartinez989:
[save_FacilityInfiltrate.zip](https://github.com/long-war-2/lwotc/files/3126251/save_FacilityInfiltrate.zip)
It looks like there is a problem with trying to infiltrate an alien facility. If you uncover an alien facility and then try to infiltrate it there is no way to check the infiltration status on it, the label for the infiltration status never shows up. When you click on the facility again it brings up the "Assualt Facility" menu again acting as if it does not have a squad infiltrating it. If you send another squad to infiltrate while one is already there then it will just give you the option to abort the first attempted infiltration.

Attached my save file of this.

pledbrook:
So the problem here is that the Geoscape icon for facilities is of type `UIStrategyMapItem_AlienNetworkComponent`, whereas all the of the other mission sites are represented by `UIStrategyMapItem_Mission`, which we override with `UIStrategyMapItem_Mission_LW` ([see _XComEngine.ini_](https://github.com/long-war-2/lwotc/blob/master/LongWarOfTheChosen/Config/XComEngine.ini#L42)).

The only solution I can think of based on my limited knowledge is the following:

 1. Create a `UIStrategyMapItem_AlienNetworkComponent_LW` class that extends the vanilla version
 2. Set up a ModClassOverride for this in _XComEngine.ini_
 3. Factor out the method implementations of `UIStrategyMapItem_Mission_LW` into a separate, reusable class. You might need to put implement the shared logic as static methods. I don't know. You certainly can't put them in a shared super-class.
 4. Add the set of methods defined in `UIStrategyMapItem_Mission_LW` to `UIStrategyMapItem_AlienNetworkComponent_LW` and have the two sets of methods (one in each class) use the shared implementations.

It sounds like it should be fairly straightforward, but I'm sure the devil is in the detail.

jmartinez989:
Ok this actually gives me a lot to work with. I'll see how much progress I can make with this.

jmartinez989:
Ok so I got it to the point where this is almost working as would be expected. I can now send a squad to infiltrate and the progress bar shows up with the infiltration percentage. Once it is at 100 percent I can start infiltration and mission starts without a problem and plays like how an alien facility normally would.

Can place x-4 charge and leave and complete mission. The only thing that is left is that when the mission is done there is no notification that avatar project progress has been reduced (the amount of avatar project is reduced by the number of doom pips the facility had).

[...]

PR #163, opened 2019-05-11
Unable to infiltrate alien facility fix (#139)

jmartinez989:
This push is to fix the issue of not being able to infiltrate alien
facilities (Issue #139). You can now infiltrate a facility and complete
it successfully. It will remove the amount of doom the facility has but
will not play the animation and display the notification of having done
so. This isa WIP commit and will need to figure out the hook to call to
play the animation.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIStrategyMapItem_AlienNetworkComponent_LW.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIStrategyMapItem_AlienNetworkComponent_LW.uc
@@ -1,0 +1,347 @@
+//---------------------------------------------------------------------------------------
+//  FILE:    UIStrategyMapItem_AlienNetworkComponent_LW
+//  AUTHOR:  Joey Martinez -- 05/06/2019
+//  PURPOSE: This file represents an alien network spot on the StrategyMap for Long War of the Chosen.
+//--------------------------------------------------------------------------------------- 
+
+class UIStrategyMapItem_AlienNetworkComponent_LW extends UIStrategyMapItem_AlienNetworkComponent;
+
+var UIScanButton ScanButton;  // used for missions being infiltrated
+var UIText InfilPct;
+var UIText InfilLabel;
+
+var UIImage SquadImage; // for cases where there is an infiltrating squad
+var string CachedImageName;
+
+//var UIStrategyMap_MissionIcon_LW MissionSiteIcon; // used for pending missions not yet infiltrated 
+var UIProgressBar ProgressBar; // used to show remaining time for missions not yet infiltrated
+
+var int InDepth; // tracks depth of mouse in and outs, since it's recording from two possibly overlapping elements (2D and 3D icons)
+
+var localized string strRecon;
+
+var transient bool bScanButtonResized;
+var transient float CachedScanButtonWidth;
+
+simulated function UIStrategyMapItem InitMapItem(out XComGameState_GeoscapeEntity Entity)
+{
+	// override the super so we can always use MI_alienfacility to allow displaying of doom pips
+	//super.InitMapItem(Entity);
+
+	local string PinImage;
+	// Initialize static data
+	InitPanel(Name(Entity.GetUIWidgetName()), 'MI_alienFacility');
+
+	PinImage = Entity.GetUIPinImagePath();
+	if( PinImage != "" )
+	{
+		SetImage(PinImage);
+	}
+
+	InitFromGeoscapeEntity(Entity);
+
+	ScanButton = Spawn(class'UIScanButton', self).InitScanButton();
+	ScanButton.SetButtonIcon("");
+	ScanButton.SetDefaultDelegate(OpenInfiltrationMissionScreen);
+	ScanButton.SetButtonType(eUIScanButtonType_Default);
+
+	SquadImage = Spawn(class'UIImage', self);
+	SquadImage.bAnimateOnInit = false;
+	SquadImage.InitImage().SetSize(48, 48);
+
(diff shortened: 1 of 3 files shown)


LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIStrategyMapItem_AlienNetworkComponent_LW.uc
```diff
@@ -0,0 +1,347 @@
...
+class UIStrategyMapItem_AlienNetworkComponent_LW extends UIStrategyMapItem_AlienNetworkComponent;
+
+var UIScanButton ScanButton;  // used for missions being infiltrated
+var UIText InfilPct;
+var UIText InfilLabel;
+
+var UIImage SquadImage; // for cases where there is an infiltrating squad
+var string CachedImageName;
+
+//var UIStrategyMap_MissionIcon_LW MissionSiteIcon; // used for pending missions not yet infiltrated
```
pledbrook: Is this needed or not? If it's not, then remove it. It's confusing to have commented out code without an explanation as to why it's commented out and whether it may be uncommented at a later date.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIStrategyMapItem_AlienNetworkComponent_LW.uc
```diff
@@ -0,0 +1,347 @@
+//---------------------------------------------------------------------------------------
+//  FILE:    UIStrategyMapItem_AlienNetworkComponent_LW
+//  AUTHOR:  Joey Martinez -- 05/06/2019
+//  PURPOSE: This file represents an alien network spot on the StrategyMap for Long War of the Chosen.
```
pledbrook: I would consider adding an explanation as to why we're overriding the base game class with this.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIStrategyMapItem_AlienNetworkComponent_LW.uc
```diff
@@ -0,0 +1,347 @@
...
+}
+
+function OpenInfiltrationMissionScreen()
+{
+	local UIMission_LWLaunchDelayedMission MissionScreen;
+	local XComHQPresentationLayer HQPres;
+	
+	// WOTC DEBUGGING
+	`LWTrace(">> Launching delayed mission UI");
+	// END
```
pledbrook: I would remove this if it's no longer needed for debugging.

jmartinez989:
> Thanks for doing this. Before merging, though, could you mark the changes from the base game class? And might it be better as a subclass of `UIStrategyMapItem_AlienNetworkComponent`?

Mark the changes as in put in comments describing the changes? The entire file is just a literal copy paste of UIStrategyMapItem_Mission_LW so that the alien facility can be treated as a mission to infiltrate. I can put that in the comment at the top of the file.

pledbrook:
Sorry, I should have been clear. Basically, put the code that is different from the original file within comment blocks, as is done with the community highlander. For example:

```
// LWOTC: Changed this to do something because of some reason
<new/changed code>
// End
```

jmartinez989:
> Sorry, I should have been clear. Basically, put the code that is different from the original file within comment blocks, as is done with the community highlander. For example:
> 
> ```
> // LWOTC: Changed this to do something because of some reason
> <new/changed code>
> // End
> ```

Ahh well that one's easy. The entire file is new code with regard to the parent class. There really is no reference to the parent class in it. All of it is new behavior. I'll just state so at the top of the file if that works.

The base class implementation is actually super short, see below:

```
//  *********   FIRAXIS SOURCE CODE   ******************
//  FILE:    UIStrategyMapItem_AlienNetworkComponent
//  AUTHOR:  Mark Nauta -- 01/12/2015
//  PURPOSE: This file represents an alien network spot on the StrategyMap.
//  Copyright (c) 2016 Firaxis Games, Inc. All rights reserved.

class UIStrategyMapItem_AlienNetworkComponent extends UIStrategyMapItem;

function UpdateFlyoverText()
{
	local XComGameStateHistory History;
	local XComGameState_MissionSite MissionState;

	History = `XCOMHISTORY;
	MissionState = XComGameState_MissionSite(History.GetGameStateForObjectID(GeoscapeEntityRef.ObjectID));
	SetLevel(MissionState.Doom);
}

simulated function bool IsSelectable()
{
	return true;
}

defaultproperties
{
}
```

pledbrook:
Oh I see. Then mention that it's a copy of `UIStrategyMapItem_Mission_LW` and point out any differences from that class.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIStrategyMapItem_AlienNetworkComponent_LW.uc
jmartinez989: done

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIStrategyMapItem_AlienNetworkComponent_LW.uc
jmartinez989: done

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIStrategyMapItem_AlienNetworkComponent_LW.uc
jmartinez989: done

Issue #140, opened 2019-04-29
Baseline enemy activity wrong on all missions when shadow chamber built


jmartinez989:
When you build the shadow chamber the baseline enemy activity for any mission that spawns will always be extremely light, even in areas where advent strength is 5 or higher. I didn't bother to check if that is actually how many enemies the mission spawns but I am willing to bet that it spawns the amount of enemies that is correct for the actual advent strength in that area.

Will try to get a save file of this as I accidentally deleted the one I had with  a shadow chamber.

jmartinez989:
Update: so this appears to not only be a visual bug. The shadow chamber literally sets all mission baseline activity to extremely light, regardless of advent strength. This makes advent strength irrelevant.

jmartinez989:
Update number 2: So this goes so far as to make the baseline activity extremely light no matter how much you have infiltrated a mission. Meaning you could start infiltration on a mission and start it right away and it would still be extremely light.

This does not seem to be affecting things like alien facilities or golden path mission sites. Seems to only be missions spawned from scanning.

jmartinez989:
Update 3: So it seems the at least the other things that are affected by advent strength like how early reinforcements com and how many reinforcements come is not affected. Also the sky ranger does take longer to come still if you start a mission way earlier than you should so it is still recommended to infiltrate.

jmartinez989:
Update 4: So if boost infiltration on a mission it will change the enemy baselines activity to whatever it would be at the percentage it is at relative to the advent strength in the area but the shadow chamber will stills that there are only 8 or 9 enemies and then when you try to launch it the game crashes.

PR #172, opened 2019-05-18
Shadow Chamber Issues Fix (#140)

jmartinez989:
The Shadow Chamber had 2 issues. The first was that it displayed
information for missions that were not Golden Path missions or Alien
Facility missions which it was not supposed to do. The second issue
was that it set alert level to all missions to 1 which caused baseline
enemy activity to be wrong (too low) for a lot of the regions come late
game.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIMission_LWCustomMission.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIMission_LWCustomMission.uc
@@ -138,79 +138,11 @@ simulated function BindLibraryItem()
 
 			break;
 		default:
-			/*********************************** Issue #140 ***********************************
-			* The code for this default case replaces the call to super.BindLibraryItem(). It's actually just a copy/paste
-			* of the super.BindLibraryItem() function with the below 2 lines taken out:
-			*     ShadowChamber = Spawn(class'UIAlertShadowChamberPanel', LibraryPanel);
-			*     ShadowChamber.InitPanel('UIAlertShadowChamberPanel', 'Alert_ShadowChamber');
-			*
-			* This is done so that normal missionss (gops, suply raids etc..) do not show shadow chamber information as
-			* this behavior is what vanilla LW2 implemented.
-		    **********************************************************************************/
-			AlertLibID = GetLibraryID();
+			super.BindLibraryItem();
 
-			if( AlertLibID != '' )
-			{
-				LibraryPanel = Spawn(class'UIPanel', self);
-				LibraryPanel.bAnimateOnInit = false;
-				LibraryPanel.InitPanel('LibraryPanel', AlertLibID);
-		
-				ButtonGroup = Spawn(class'UIPanel', LibraryPanel);
-				ButtonGroup.InitPanel('ButtonGroup', '');
-		
-				Button1 = Spawn(class'UIButton', ButtonGroup);
-				Button2 = Spawn(class'UIButton', ButtonGroup);
-		
-				Button3 = Spawn(class'UIButton', ButtonGroup);
-				Button3.SetResizeToText( false );
-				Button3.InitButton('Button2', "",, eUIButtonStyle_NONE);
-				Button3.OnSizeRealized = OnButtonSizeRealized;
-		
-				ConfirmButton = Spawn(class'UIButton', LibraryPanel);
-		
-				if( `ISCONTROLLERACTIVE == false )
-				{
-					ConfirmButton.InitButton('ConfirmButton', "", OnLaunchClicked);
-					Button1.InitButton('Button0', "",, eUIButtonStyle_NONE);
-					Button2.InitButton('Button1', "",, eUIButtonStyle_NONE);
-					ConfirmButton.SetResizeToText(false);
-					Button1.SetResizeToText( false );
-					Button2.SetResizeToText(false);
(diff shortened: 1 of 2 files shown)


LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIMission_LWCustomMission.uc
```diff
@@ -135,10 +135,83 @@ simulated function BindLibraryItem()
...
 				ShadowChamber.InitPanel('ShadowChamber');
 			}
+
 			break;
 		default:
-			super.BindLibraryItem();
-			break;
+			/*********************************** Issue #140 ***********************************
+			* The code for this default case replaces the call to super.BindLibraryItem(). It's actually just a copy/paste
+			* of the super.BindLibraryItem() function with the below 2 lines taken out:
```
pledbrook: Could we not just call the `super` function here and hide the panel afterwards?

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIMission_LWCustomMission.uc
```diff
@@ -344,7 +418,24 @@ simulated function UpdateData()
...
-	super.UpdateData();
+	// Region Panel
+	if( LibraryPanel == none )
+	{
+		UpdateTitle('Region', GetRegion().GetMyTemplate().DisplayName, GetLabelColor(), 50);
+	}
+	
+	/*********************************** Issue #140 ***********************************
+	* THe below code is replacing the call to super.UpdateData(). The code from the parent
+	* class verion of UpdateData() is pasted here except the call to UpdateMissionSchedules()
```
pledbrook: Typos, but otherwise, great. It's good to have informative comments like this that explain what's been done and why.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIMission_LWLaunchDelayedMission.uc
```diff
@@ -52,6 +57,138 @@ simulated function Name GetLibraryID()
...
+				SitrepPanel.SetTitle(m_strSitrepTitle);
+		
+				ChosenPanel = Spawn(class'UIPanel', LibraryPanel).InitPanel(, 'Alert_ChosenRegionInfo');
+				ChosenPanel.DisableNavigation();
+			}
+
+		    break;
+	}
+}
+
```
pledbrook: As with the earlier commit, can't we just call the super method and then hide the shadow chamber panel?

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIMission_LWCustomMission.uc
jmartinez989: I'm such a derp. I fixed this with my latest commit. Just put the call to super.BindLibraryItem() back in and then just called ShadowChamber.Hide() in there to hide the panel.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIMission_LWCustomMission.uc
jmartinez989: Corrected the typos in the latest commit.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIMission_LWLaunchDelayedMission.uc
jmartinez989: resolved with my latest commit.

pledbrook: Is the rest of the copied code necessary here, i.e. what was copied from `UIMission_LWCustomMission`? Couldn't the entire implementation of this method simply be:

```
super.BindLibraryItem();

ShadowChamber.Hide();
```

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIMission_LWLaunchDelayedMission.uc
```diff
@@ -67,53 +67,19 @@ simulated function BindLibraryItem()
...
-			super.BindLibraryItem();
-
+		case eMissionUI_GuerrillaOps:
+		case eMissionUI_SupplyRaid:
+		case eMissionUI_LandedUFO:
+		case eMissionUI_GPIntel:
+		case eMissionUI_Council:
+		case eMissionUI_Retaliation:
+        case eMissionUI_Rendezvous:
+		case eMissionUI_Invasion:
```
pledbrook: You could also have a simple `if` condition, e.g. `if (MissionUIType != eMissionUI_GoldenPath)`. That will also scale better if we add new mission types.

Also, is the shadow chamber panel supposed to be visible for facility missions?

jmartinez989: So as far as facility goes, it is never visible even if I leave the check out of it in the code. Not sure what is going on there but it is never visible not matter what I do. Guess I can just  turn it into an if for Golden Path though. Will do that then.

jmartinez989: ok. changed from a switch to an if.

PR #202, opened 2019-06-03
Disable reusable weapon upgrades breakthrough

### LINKED ISSUE #179: Disable Interchangable Upgrades Breakthrough
The functionality of Interchangeable Upgrades is included in LW as standard, so having it as a Breakthrough is redundant.

Kazanir:
Make the reusable weapon upgrades tech unresearchable as it is a base feature in LW. Fixes #179.

### DIFF
--- a/LongWarOfTheChosen/Config/XComLW_Overhaul.ini
+++ b/LongWarOfTheChosen/Config/XComLW_Overhaul.ini
@@ -763,6 +763,9 @@ INSTANT_BUILD_TIMES=TRUE
 ALIEN_FACILITY_LEAD_RP_INCREMENT=1200
 ALIEN_FACILITY_LEAD_INTEL=5
 
+;Disable redundant WOTC techs/breakthroughs
++TechTable=(TechTemplateName="BreakthroughReuseWeaponUpgrades",	ProvingGround=false,	ModPointsToCompleteOnly=False,	PrereqTech1="", PrereqTech2="",	PrereqTech3="",	RequiredScienceScore=99999)
+
 ; Repeatedable permanent improvement projects; the increased cost for repeating is in DGD.ini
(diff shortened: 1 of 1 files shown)


LongWarOfTheChosen/Config/XComLW_Overhaul.ini
```diff
@@ -763,6 +763,9 @@ INSTANT_BUILD_TIMES=TRUE
 ALIEN_FACILITY_LEAD_RP_INCREMENT=1200
 ALIEN_FACILITY_LEAD_INTEL=5
 
+;Disable redundant WOTC techs/breakthroughs
++TechTable=(TechTemplateName="BreakthroughReuseWeaponUpgrades",	ProvingGround=false,	ModPointsToCompleteOnly=False,	PrereqTech1="", PrereqTech2="",	PrereqTech3="",	RequiredScienceScore=99999)
```
jmartinez989: So does this actually disable the breakthrough from showing up or does it just make it super long to research to the point where trying to do so is pointless? If it is the latter then we would want it to be changed so that it is removed from the deck of breakthroughs so that it does not occupy the spot of another possible breakthrough.

micahswab: I believe that represents scientist requirements. Since you’ll never have enough scientists to reach that science score, it effectively removes the tech.

Kazanir: I copied the way the other techs are disabled lower down in the ini file; here's hoping that the breakthrough selection process respects that the tech is not available

pledbrook: This is indeed the standard way to "remove" techs from the game. Your scientists will never accumulate enough research points, so the tech can never be researched.

There is also a boolean flag on tech templates that determines whether the tech should be displayed regardless of whether it can be researched or not. By default, it's `false`, so tech that can't be researched is hidden. (Or the flag is the other way round and the default value is `true` - can't remember)

PR #205, opened 2019-06-03
Fix SPARK tech upgrades

### LINKED ISSUE #203: SPARK tech tree issues
It seems as if:

- The MechanizedWarfare tech is not properly available to research in the PG
- SPARKs can be built without the tech being researched
- This leads to having SPARKs whose gear cannot be upgraded
- I also don't have the BIT bombing ability on my SPARKs; may or may not be related

Kazanir:
Fix SPARK tech upgrades via WOTC integrated DLC helper class instead of LW2DLCHelper

Fixes #203.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/LWDLCHelpers.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/LWDLCHelpers.uc
@@ -137,6 +137,21 @@ static function bool IsLostTowersNarrativeEnabled()
 	return false;
 }
 
+// Replaces X2StrategyElement_DLC_Day90Techs.static.IsMechanizedWarfareAvailable()
+static function bool IsMechanizedWarfareAvailable()
+{
+    // We need MechanizedWarfare to be available in Integrated DLC mode. This function replaces the
+	// default SpecialRequirementsFn for MechanizedWarfare from WOTC with our own, making the tree
+	// function in the expected LW2 style. See also LWTemplateMods.uc for the actual replacement.
+	local XComGameState_CampaignSettings CampaignSettings;
+
+	CampaignSettings = XComGameState_CampaignSettings(`XCOMHISTORY.GetSingleGameStateObjectForClass(class'XComGameState_CampaignSettings'));
+
+	if (CampaignSettings != none)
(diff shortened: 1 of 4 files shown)


Kazanir:
Ok, I believe this will work correctly.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/LWDLCHelpers.uc
```diff
@@ -137,6 +137,21 @@ static function bool IsLostTowersNarrativeEnabled()
 	return false;
 }
 
+// Replaces X2StrategyElement_DLC_Day90Techs.static.IsMechanizedWarfareAvailable()
+static function bool IsMechanizedWarfareAvailable()
+{
+    // We need MechanizedWarfare to be available in Integrated DLC mode. This function replaces the
```
pledbrook: I know it's a trivial thing, but this is going to drive me nuts. A tab here please.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/LWTemplateMods.uc
```diff
@@ -2689,6 +2689,12 @@ function RewireTechTree(X2StrategyElementTemplate Template, int Difficulty)
 		if (TechTemplate.DataName == 'HeavyWeapons') // remove the alternative access to the heavy weapons proving ground project for sparks
 			TechTemplate.AlternateRequirements.Length = 0;
 
+		// Change the special requirements functions for Mechanized Warfare & SPARKs.
+		if (TechTemplate.DataName == 'MechanizedWarfare')
+			TechTemplate.Requirements.SpecialRequirementsFn = class'LWDLCHelpers'.static.IsMechanizedWarfareAvailable;
```
pledbrook: Are you sure about this? Shouldn't the special requirements function be the same as for BuildSpark, i.e. has Lost Towers been completed?

pledbrook: I think this is the right approach, i.e. replacing the SpecialRequirementsFn, I'm just not sure this is the right function. I'm pretty sure the point is for the SpecialRequirementsFn to override the prerequisites, which is ADVENT Robotics for Mechanized Warfare.

pledbrook:
Thanks. Confirmed it works for the three scenarios I mentioned.

Issue #100, opened 2019-04-16
Avatar progress doesn't seem to be progressing properly


pledbrook:
I've had reports of the doom clock acting oddly, for example with every added pip being removed again. This removes the doom clock from being a consideration at all for campaigns.

jmartinez989:
So I found that the Avatar project clock is actually going up like it is supposed to but you only see that if you actually uncover the facilities that have been adding to the avatar project. If they are not revealed then the clock will add a pip then remove one. Once you uncover that facility you will see all of the pips correctly that it added.

pledbrook:
That's weird. Wonder what's going on. I need a pre-facility-reveal save file if anyone has one lying around.

jmartinez989:
Look at #139 That has a save file for another issue with alien facilities. That save I have revealed a facility.

pledbrook:
Right. I was hoping for one just before the facility is revealed :smile: A tough ask, I know. I can probably work out what's going on from other saves, like that one and HQ assaults.

jmartinez989:
Have a solution implemented for this but part of it requires a CHL modification so will need to push that change to the highlander repo as well.

PR #231, opened 2019-06-15
Add event listener for avatar project progress

jmartinez989:
Add an event listener that utilizes an event hook that listens for an
event fired whenever doom is updated. The event listener adds up all
of the doom for any facilities that have not been revealed and adds
that to all of the doom that has currently been gathered so that the
amount of doom is updated properly. The issue before this was that
alien facilities that were not revealed had an issue when updating doom
in that when they added a point of doom that point would be removed
then added which would make it look like it did not move up at all but
then when the facility was revealed it would add up all of the doom
that the facility had actually added.


pledbrook:
Don't forget to include issue numbers in your commit messages when they are relevant. In this case, a simple "Fixes #100" line would be sufficient. I'll make the change with the merge.

jmartinez989:
> Don't forget to include issue numbers in your commit messages when they are relevant. In this case, a simple "Fixes #100" line would be sufficient. I'll make the change with the merge.

Thanks, sorry I forgot. I'll make sure to do that for the rest.

PR #271, opened 2019-06-23
Limit Death from Above to only refunding a single action per turn

Kazanir:


pledbrook:
Thanks. As a small favour, could you follow commit message conventions? Primarily making the first line a short summary (max 55 chars or so), followed by a blank line and then further description of the changes as max 80 character lines.

Also, please create issues for user-visible changes like this so that they can automatically be included in the milestone. Users can then easily see what's changed for any given release and it makes it easier to track changes for release notes.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2Effect_DeathFromAbove_LW.uc
```diff
@@ -8,12 +8,23 @@ function bool PostAbilityCostPaid(XComGameState_Effect EffectState, XComGameStat
...
     if(SourceUnit.IsUnitAffectedByEffectName(class'X2Effect_Serial'.default.EffectName))
     {
         return false;
     }
 
+	// Block repeated uses of DFA by default.
+	// CREDIT TO: Favid ([WOTC] LW2 Classes and Perks)
+	SourceUnit.GetUnitValue('LW_DeathFromAboveUses', DFAUsesThisTurn);
+	iUsesThisTurn = int(DFAUsesThisTurn.fValue);
+	if (iUsesThisTurn >= default.DFA_USES_PER_TURN)
```
pledbrook: I think you forgot to add the config var to this class! I'll fix.

Issue #251, opened 2019-06-19
Empty UI window after completing network tower mission


pledbrook:
[save_AUTOSAVE-_Campaign_43_Mission_137_Turn_16.zip](https://github.com/long-war-2/lwotc/files/3307737/save_AUTOSAVE-_Campaign_43_Mission_137_Turn_16.zip)

BlackRose39:
Right, I am pretty sure that that is suppose to be the screen indicating that the HQ assault is ready, and is prompting the player to view/wait to launch.

PR #272, opened 2019-06-23
Council Mission alert screen for HQ Assault

Kazanir:
Not sure what LW2 was doing here since it is also missing this bit of config and gives a Council Message alert with only one button. But this will fix the immediate issue in #251.


pledbrook:
The reason I haven't merged this is because the save file attached to the corresponding issue crashes. So I haven't been able to test either the current behaviour or the fixed one.

pledbrook:
I will get round to this, honest!

PR #304, opened 2019-07-05
Jm/issue 121 fix

jmartinez989:
Soldier defense was not being displayed in the armory like it was in
lw2. This uses a hook into a CHL event that allows overriding the
soldier header in the armory to display the defense stat.

The defense stat will be displayed at the bottom right corner of the
header where the psi stat normally goes unless the soldier is a psi
operative. In that case the defense stat replaces the hack stat and
the psi stat goes where it normally does.


pledbrook:
The eventual changes to the highlander are unlikely to support this. Will create a new PR when the highlander stuff is sorted.

Issue #341, opened 2019-07-15
missing hunter's unlockable pistol? 


captainbeastfeast:
I unlocked the chosen in ini and beat them all, researched their weapons. I just realised that I don't have access to the hunter's pistol at all. His rifle is available but his pistol isn't available in my utility items and never has been since it was unlocked months ago.

pledbrook:
@jmartinez989 Does this need to be made a buildable item? This is confusing because the in-code name of the Shadowkeeper is HunterPistol or HuntersPistol. Don't know what the internal name of the Hunter's pistol is.

captainbeastfeast:
the official name of his pistol is the 'Darkclaw.' it does 4-7 damage in the base game and ignores 5 points of armour. See wiki:
https://xcom.fandom.com/wiki/Darkclaw
I don't know what the name of it in game code is, but it is a nice pistol to have, especially for snipers who have pistol wielding skills :)

The pistol doesn't need to be made build-able, but a single pistol should be made available after completing chosen weapons research (like all the other chosen weapons). Hope this helps, thanks.

------

P.S on the issue of chosen, please consider making their ballistic weapons do rupture (1) like final tier  alien weaponry in LW2.  This isn't needed for the assassin's katana as it is already very strong and it is a melee weapon. It would be fine balance wise if these effects apply to chosen when they use their own weapons, as well as when xcom use them.

jmartinez989:
as @captainbeastfeast stated it does not need to be buildable. It just needs to be granted after completing the research of the chosen hunter weapons. I'll take this one. I have quite a back log of things I would like to work on so I will add this on there.

jmartinez989:
This is actually pretty easy a fix. Just need to change the inventory slot type of the pistol to `eInvSlot_Utility`.

jmartinez989:
I implemented a fix to this that turns the dark claw into a utility pistol. Will push the changes up to a new branch and create a merge request.

jmartinez989:
pushed the fix up and made a pull request. just waiting for approval so closing this for now.

PR #345, opened 2019-07-16
Make Darkclaw a Utility Item (#341)

jmartinez989:
Chosen Hunter pistol was not classified as a utility slot pistol. Added
the needed code in LWTemplateMods to change it. It does the same thing
that is done for the alien hunter pistol.


pledbrook:
Thanks!

Issue #317, opened 2019-07-09
SMG's getting bonus from Assault Rifle damage breakthroughs


jmartinez989:
The reason this is happening is because SMG weapon categories are the same as assault rifle. This just needs the proper changes in the smg classes and some ini files.

jmartinez989:
Changed the weapon category of smg to 'smg' from 'rifle' and now do not see the smgs getting the bonus from the assault rifle breakthrough. Closing this as now all that is left is to merge a pull request.

PR #346, opened 2019-07-16
Changed SMG Weapon Category (#317)

jmartinez989:
SMGs were getting the damage bonus from the assault rifle weapon
breakthrough. This was because their weapon category was 'rifle' which
is the same category as the assualt rifle. Changed the weapon category
to 'smg' and added the needed lines in XComClassData.ini to allow the
classes that could equipem them to continue to equip them.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_LaserPack_Integrated/Classes/X2Item_LaserWeapons.uc
+++ b/LongWarOfTheChosen/Src/LW_LaserPack_Integrated/Classes/X2Item_LaserWeapons.uc
@@ -6,8 +6,8 @@
 //---------------------------------------------------------------------------------------
 class X2Item_LaserWeapons extends X2Item config(GameData_WeaponData);
 
-// Variables from config - GameData_WeaponData.ini
-// ***** Damage arrays for attack actions  *****
+//ï¿½Variablesï¿½fromï¿½configï¿½-ï¿½GameData_WeaponData.ini
+//ï¿½*****ï¿½Damageï¿½arraysï¿½forï¿½attackï¿½actionsï¿½ï¿½*****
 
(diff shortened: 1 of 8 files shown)


LongWarOfTheChosen/Src/LW_LaserPack_Integrated/Classes/X2Item_LaserWeapons.uc
```diff
@@ -6,8 +6,8 @@
 //---------------------------------------------------------------------------------------
 class X2Item_LaserWeapons extends X2Item config(GameData_WeaponData);
 
-//�Variables�from�config�-�GameData_WeaponData.ini
-//�*****�Damage�arrays�for�attack�actions��*****
+//�Variables�from�config�-�GameData_WeaponData.ini
+//�*****�Damage�arrays�for�attack�actions��*****
```
pledbrook: What's this?

jmartinez989: I do not know how this happened. Somehow my editor replaced the spaces with unprintable characters.

I'll get this fixed.

jmartinez989: This is fixed now. I'll make sure to keep an eye for this again.

WilleBack:
This does mean that there is now no Weapon Breakthrough for the SMG, so one should probably be created if the breakthroughs are to be retained.

jmartinez989:
> This does mean that there is now no Weapon Breakthrough for the SMG, so one should probably be created if the breakthroughs are to be retained.

We do not yet even know if we are keeping any of the breakthroughs. Maybe the tier specific breakthroughs we will but the weapon type breakthroughs might receive the axe.

PR #453, opened 2019-08-28
Allow rapidfire cooldown to be set in config

s4320079:
Looks like WotC nerfed the base rapid fire, possibly because the cooldownless version broke the balance of other classes when given to them with the AWS. I've implemented a replacement that has a cooldown that can be changed in config and replaced all occurrences of the old version.


pledbrook:
Thanks for looking into this. Why not simply modify the existing Rapid Fire template to remove or reduce its cooldown?

s4320079:
Rapid fire is a vanilla ability and I heard it was good practice to replace rather than alter vanilla code where possible. Even if a lot of the reasons for that aren't relevent here (ie. Highlander means mod compatibility is hard and firaxis isn't altering files anymore) I still think there is value in separating mod and vanilla code as much as possible.

pledbrook:
The downside of this is that any player that wants to use another soldier class that has Rapid Fire in its tree will get the 5-turn cooldown.

> and I heard it was good practice to replace rather than alter vanilla code where possible

But what is the specific reason to do so in this case? If we were to fundamentally change the behaviour of Rapid Fire, I could understand. But it's still a double shot with an aim malus. The only change is cooldown.

s4320079:
I changed the ability used by all long war classes to the new one and if someone adds their own class and gives it the vanilla rapidfire I would assume they wanted the vanilla rapidfire for balance reasons. 

I consider adding another soldier class with rapidfire and getting the 5 turn cooldown version to be the upside in this case. Stops someone else's class mod from acting differently depending on if Lwotc is installed. 

Plus from what I can see Ranger_Ability_set doesn't look to be altered in LW yet and adding an extra function to the perk pack is a smaller change than adding a whole new file.

pledbrook:
Allowing two abilities with the same name that behave differently is a non starter. There is no good reason _for an overhaul mod_ not to simply change the cooldown on the base game's Rapid Fire ability. Otherwise you would really need to add a new ability.

See _LWTemplateMods.uc_ for several examples of LW2 modifying existing abilities. For example, it adjusts the cooldowns for the Spider Suit and Wraith Suit grapples.

s4320079:
Woops I see what you mean now. Made the change.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/LWTemplateMods.uc
```diff
@@ -1176,6 +1177,12 @@ function ModifyAbilitiesGeneral(X2AbilityTemplate Template, int Difficulty)
 		}
 	}
 
+	if(Template.DataName == 'RapidFire')
```
pledbrook: For future reference, please put spaces after `if`, `for`, `switch`, etc. I know there are plenty of inconsistencies in the current code, but want to make new stuff follow a standard. I'll fix it locally.

Issue #461, opened 2019-08-31
Shaken soldiers can become haven advisers


pledbrook:
This makes no sense as Shaken soldiers are supposed to require time on the Avenger to recover. Haven advisers are not on the Avenger.

See my SSGT Grenadier in this save, who is currently Shaken and an adviser:

[save_BUG Shaken soldiers as haven advisers.zip](https://github.com/long-war-2/lwotc/files/3562459/save_BUG.Shaken.soldiers.as.haven.advisers.zip)

ansabalirai:
I am perfectly fine with the change, but just wondering whether this is different from vanilla LW2? I think I remember being able to station shaken soldiers in havens. Not sure if they recovered while being an advisor since vanilla required the soldier to return uninjured and having gotten a kill in a subsequent mission to become unshaken.

pledbrook:
You couldn't have Shaken soldiers in original LW2, could you? I thought it's a WOTC thing with mental states.

PR #462, opened 2019-08-31
#461 Shaken soldiers can no longer serve as haven advisers

Favid:
- Can no longer select shaken soldiers to be advisers
- If an adviser becomes shaken on/after a mission, they will be removed as an adviser
- Shaken and Rank Too Low soldier statuses now have higher priority than Unit in Squad when on the select adviser screen

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2EventListener_Soldiers.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2EventListener_Soldiers.uc
@@ -251,12 +251,9 @@ static protected function EventListenerReturn OnOverridePersonnelStatus(Object E
 		{
 			SetStatusTupleData(OverrideTuple, class'UIUtilities_Strategy'.default.m_strOnMissionStatus, "", "", 0, eUIState_Highlight, true);
 		}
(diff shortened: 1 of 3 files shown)


LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2EventListener_Soldiers.uc
```diff
@@ -251,11 +251,18 @@ static protected function EventListenerReturn OnOverridePersonnelStatus(Object E
...
+			{
+				SetStatusTupleData(OverrideTuple, default.RankTooLow, "", "", 0, eUIState_Bad, true);
+			}
+		}
 		else if (SquadMgr != none && SquadMgr.UnitIsInAnySquad(UnitState.GetReference(), Squad))
 		{
 			if (SquadMgr.LaunchingMissionSquad.ObjectID != Squad.ObjectID)
 			{
-				if (UnitState.GetStatus() != eStatus_Healing && UnitState.GetStatus() != eStatus_Training)
+				if (UnitState.GetStatus() != eStatus_Healing && UnitState.GetStatus() != eStatus_Training && UnitState.GetMentalState() != eMentalState_Shaken)
```
pledbrook: I wonder whether this might as well just check for `eStatus_Active`. Otherwise it also catches soldiers in `eStatus_PsiTesting`/`eStatus_PsiTraining`. And I don't think a soldier can be Shaken and active, though I'm not sure.

Issue #274, opened 2019-06-24
Rebels don't get AWC abilities


pledbrook:
The code for doing this is commented out in `XComGameState_LWOutpost.PromoteRebel()` because it relies on AWC stuff. Which may work as is in WOTC, but I suspect it needs some changes. And LW2's custom AWC behaviour hasn't been ported yet.

PR #470, opened 2019-09-02
#274 Rebels should now gain abilities when they rank up

Favid:


LongWarOfTheChosen/Src/LW_Overhaul/Classes/XComGameState_LWOutpost.uc
```diff
@@ -1763,3 +1801,30 @@ function bool ShouldBeVisible()
...
+{
+	local RebelUnit Rebel;
+	local XComGameState_Unit UnitState;
+
+	foreach Rebels(Rebel)
+	{
+		UnitState = XComGameState_Unit(NewGameState.CreateStateObject(class'XComGameState_Unit', Rebel.Unit.ObjectID));
+		NewGameState.AddStateObject(UnitState);
+		// Rebel has ranked up but has no abilities, so give them abilities
+		if (Rebel.Level > 0 && UnitState.GetAWCAbilityNames().Length == 0)
```
pledbrook: `GetAWCAbilityNames()` doesn't work because it has a check for `bUnlocked`, which none of the abilities has set to `true`. So this function always returns a zero-length array.

You either need to check the length of `AWCAbilities` directly, or set `bUnlocked` to true when giving each AWC perk.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/XComGameState_LWOutpost.uc
```diff
@@ -1617,57 +1625,87 @@ function PromoteRebel(StateObjectReference UnitRef, XComGameState NewGameState)
...
+	{
+		AbilityTemplate = AbilityTemplateManager.FindAbilityTemplate(PossibleAbility.AbilityName);
+		if (AbilityTemplate == none)
+			continue;
+
+		if (PossibleAbility.Level != AWCLevel)
+			continue;
+
+        if (HasEarnedAbility(UnitState, PossibleAbility.AbilityName))
+            continue;
```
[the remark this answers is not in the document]
Favid: Not a fan of continues so I'll put them all in one if statement and flip them

LongWarOfTheChosen/Src/LW_Overhaul/Classes/XComGameState_LWOutpost.uc
```diff
@@ -1617,57 +1625,87 @@ function PromoteRebel(StateObjectReference UnitRef, XComGameState NewGameState)
...
-			GiveAWCPerk(Unit, class'LWAWCUtilities'.default.AWCAbilityTree_Offense, 1); 
-			GiveAWCPerk(Unit, class'LWAWCUtilities'.default.AWCAbilityTree_Defense, 1); 
+			GiveAWCPerk(Unit, default.REBEL_AWC_ABILITIES_OFFENSE, 1); 
+			GiveAWCPerk(Unit, default.REBEL_AWC_ABILITIES_DEFENSE, 1); 
 			break;
 		case 2:
-			GiveAWCPerk(Unit, class'LWAWCUtilities'.default.AWCAbilityTree_Offense, 2); 
-			GiveAWCPerk(Unit, class'LWAWCUtilities'.default.AWCAbilityTree_Defense, 2); 
+			GiveAWCPerk(Unit, default.REBEL_AWC_ABILITIES_OFFENSE, 2); 
+			GiveAWCPerk(Unit, default.REBEL_AWC_ABILITIES_DEFENSE, 2);
```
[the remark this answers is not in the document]
Favid: Yes it can

LongWarOfTheChosen/Src/LW_Overhaul/Classes/XComGameState_LWOutpost.uc
Favid: GiveAWCPerk() already sets bUnlocked to true, so I'm not sure why this doesn't work. But you're right, this is still firing when they already have abilities. Weird.

pledbrook: Are you sure? Just looking at the last commit and it looks like the previous code was setting `bUnlocked` to `false` :wink:

pledbrook: @Favid I will merge this, but just want to check: should `bUnlocked` be set to `true`? I'm assuming yes, in which case I'll make the change.

pledbrook: Actually, ignore that. I was looking at where the ability was initialised, not where it was granted (at which point `bUnlocked` is indeed set to `true`). As with you, have no idea why I was running into the problem with `GetAWCAbilityNames()` but can confirm it's working.

pledbrook:
Thanks!

PR #557, opened 2019-09-17
Update XComGame.int Skirmisher Body Shield Fixes #549

### LINKED ISSUE #549: Small skill description issues
Templar: Changes to Pillar regarding usage is not reflected on it's info screen when promoting.

Skirmisher: Justice claims to end your turn (on the tool-tip for the button to trigger the ability. It does not end your turn, just uses 1 action, even against a non flanked enemy. Pulling an enemy too you, then shooting him, then shooting him again thanks to hit and run is thus possible. Not complaining, exactly, just saying the info should be there to let people know that.

Skirmisher: Body Shield promotion screen information does not give any information about usage: Type of action, limited uses, cool down, etc. Would be helpful to know when deciding to take it or not.

Specialist: Field Surgeon claims that downed and stabilized enemies do not gain a bonus from it. My downed Templar had 2 health in the post action screen, so I am guessing that surgeon is triggering when it shouldn't.

nightmaRok:
Update XComGame.int Skirmisher Body Shield Fixes #549

### DIFF
--- a/LongWarOfTheChosen/Localization/LW_PerkPack_Integrated/XComGame.int
+++ b/LongWarOfTheChosen/Localization/LW_PerkPack_Integrated/XComGame.int
@@ -692,7 +692,7 @@ LocFriendlyName="Body Shield"
 LocFlyOverText="Body Shield"
 LocLongDescription="A targeted enemy receives -20 aim and -50 critical chance against the soldier."
 LocHelpText="The targeted enemy receives -20 aim and -50 critical chance against the soldier."
-LocPromotionPopupText="<Bullet/> A targeted enemy receives -20 aim and -50 critical chance against the soldier.<br/>"
+LocPromotionPopupText="<Bullet/> A targeted enemy receives -20 aim and -50 critical chance against the soldier.<br/><Bullet/> Not cost an action.<br/><Bullet/> Body Shield has a 2 turn cooldown."
 
 [EmergencyLifeSupport X2AbilityTemplate]
 LocFriendlyName="Emergency Life Support"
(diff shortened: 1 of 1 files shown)


LongWarOfTheChosen/Localization/LW_PerkPack_Integrated/XComGame.int
```diff
@@ -692,7 +692,7 @@ LocFriendlyName="Body Shield"
 LocFlyOverText="Body Shield"
 LocLongDescription="A targeted enemy receives -20 aim and -50 critical chance against the soldier."
 LocHelpText="The targeted enemy receives -20 aim and -50 critical chance against the soldier."
-LocPromotionPopupText="<Bullet/> A targeted enemy receives -20 aim and -50 critical chance against the soldier.<br/>"
+LocPromotionPopupText="<Bullet/> A targeted enemy receives -20 aim and -50 critical chance against the soldier.<br/><Bullet/> Not cost an action.<br/><Bullet/> Body Shield has a 2 turn cooldown."
```
pledbrook: Rather than "Not cost an action", use "This is a free action". Also, see if `<Ability:SelfCooldown/>` works in place or '2' for the cooldown. We wouldn't then need to update the localisation files if and when the cooldown changes.

Lastly, any new localisation content or any changes to existing text should be wrapped in specific comments like this:
```
; LWOTC Needs Translation
LocPromotionPopupText=...
; End Translation
```
This makes it easy for anyone that wants to contribute translations to find what needs translating. I like to support case-sensitive search as well, so please use the same capitalisation as in the example.

Issue #559, opened 2019-09-18
Hard Target still missing flavor text on newest beta version, Sep 17


allenhu0912:
![20190917222632_1](https://user-images.githubusercontent.com/54481962/65118764-7795e480-d9b9-11e9-8113-a63f489b5b69.jpg)
IIRC it had no flavor text in LW2 either

Clibanarius:
Yeah, that's just what it was in LW2. Might just be time to ignore that unless you wanna write some flavor text up in here and someone can splice it in.

PR #589, opened 2019-09-22
Added Quote for Hard Target Dark Event #559

BlackRose39:
In the dark event window, the "Hard Target" doesn't have any text, so I added text to it. 

Hopefully I did my best acting Shen's character!

### DIFF
--- a/LongWarOfTheChosen/Localization/XComGame.int
+++ b/LongWarOfTheChosen/Localization/XComGame.int
@@ -4282,7 +4282,7 @@ Description="Some ADVENT robots and some other units gain the Damage Control abi
 [DarkEvent_HardTarget X2DarkEventTemplate]
 DisplayName="Tactical Upgrades: Hard Target"
 SummaryText="Some alien units gain the Hard Target ability, granting bonus dodge."
-QuoteText="\".\""
+QuoteText="\"The aliens' reflexes are becoming adept in our environment. We should invest in our scopes and make sure no one leaves without one.\""
 QuoteTextAuthor="- Chief Shen"
 PreMissionText="Grants some alien units the Hard Target ability."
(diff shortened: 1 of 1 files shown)


LongWarOfTheChosen/Localization/XComGame.int
```diff
@@ -4282,7 +4282,7 @@ Description="Some ADVENT robots and some other units gain the Damage Control abi
 [DarkEvent_HardTarget X2DarkEventTemplate]
 DisplayName="Tactical Upgrades: Hard Target"
 SummaryText="Some alien units gain the Hard Target ability, granting bonus dodge."
-QuoteText="\".\""
+QuoteText="\"The aliens' reflexes are becoming adept in our environment. We should invest in our scopes and make sure no one leaves without one.\""
```
pledbrook: Would you mind putting these lines around it so that translators know that they need to update the other language files?
```
; LWOTC Needs Translation
QuoteText=...
; End Translation

BlackRose39: #589 0ec80b55c813d852c99661447238f8eb83fecaae

Done! (I think) How is it?

pledbrook: Erm...can you make it exactly "; LWOTC Needs Translation"? It means folks can just search for "LWOTC Needs Translation" to find the bits that they need to work on.

pledbrook: Also, I like it to be searchable case sensitively, so capitalisation is important.

BlackRose39: c2f71de51ad12beaee5d913909e5d7fbd2c31eb7

Right, totally understandable since coding is very specific and precise. It is my first time joining in a github project to contribute, but hopefully after a few more times, I got this!

pledbrook:
Thanks!

Issue #482, opened 2019-09-05
Fortress behaviour is wonky


DreadnoughtCyberdisk:
At the moment the "fortress" perk is acting really inconsistent:
https://www.reddit.com/r/Xcom/comments/79iqbp/fortress_what_does_it_really_make_you_immune_to/

In my oppinion it should either be clarified, what "explosions" it wont protect you from (blazing pinions for example) or it should just be unified to include all kind of explosions (Wich would probably mean that these effects should deal actual explosion damage and not whatever damage it is atm).

Long War 2 / LWOTC is just too punishing to find out the hard way, and the decisionmaking can be heavily influenced by relying on that perk to protect you.

gen-martox:
This is an easy fix just need to change blazing pinions damage type in the weapondata.ini and it will need to be changed for the archon king

[XComGame.X2Item_DefaultWeapons]
-ARCHON_BLAZINGPINIONS_BASEDAMAGE=(Damage=5, Spread=1, PlusOne=50, Crit=0, Pierce=0, Shred=0, Tag = "", DamageType="BlazingPinions")
+ARCHON_BLAZINGPINIONS_BASEDAMAGE=(Damage=5, Spread=1, PlusOne=50, Crit=0, Pierce=0, Shred=0, Tag = "", DamageType="Explosion")

[DLC_2.X2Item_DLC_Day60Weapons]
-ARCHONKING_BLAZINGPINIONS_BASEDAMAGE=(Damage=4, Spread=2, PlusOne=0, Crit=0, Pierce=0, Shred=0, Tag = "", DamageType="BlazingPinions")
+ARCHONKING_BLAZINGPINIONS_BASEDAMAGE=(Damage=4, Spread=2, PlusOne=0, Crit=0, Pierce=0, Shred=0, Tag = "", DamageType="Explosion")

pledbrook:
Hmmm...but what are the side effects of that? I assume there's a reason that it got a dedicated damage type.

pledbrook:
The LW2 folks had Incoming! specifically treat BlazingPinions as explosive. We might be able to modify Fortress to do the same.

gen-martox:
ther shouldn’t be any side effects, I pulled that from a mod I used for wotc and I didn’t see any issues with it. The only reason I can think of it haveing  dedicate damage type is so fortress doesn’t prevent the damage

PR #611, opened 2019-09-25
Classify Blazing Pinions as Explosive Type #482

BlackRose39:
#482 

There has been confusion regarding the "Fortress" perk and its behavior regarding what it should and should not protect against. Given that the Blazing Pinions looks explosive in the game, I made this change to resolve any frustrating ambiguities during game play.

Referring to the second link in the issue, I did not make any changes to plasma or shredder blasters since those don't look like "explosions." (Kinda like the Sectopod wrath cannon)


pledbrook:
I'm nervous about this change because it could have unintended side effects. Why have a dedicated Blazing Pinions damage type otherwise?

pledbrook:
For reference, the settings for "Explosion" are:
```
Template.bCauseFracture = true;
Template.MaxFireCount = 3;
Template.MinFireCount = 1;
Template.FireChance = 100;
Template.bAllowAnimatedDeath = true;
```
and for Blazing Pinions they're:
```
Template.bCauseFracture = true;
Template.MaxFireCount = 3;
Template.MinFireCount = 1;
Template.FireChance = 50;
Template.bAllowAnimatedDeath = true;
```
So it seems the only real difference is the chance to cause fire.

MrNiceUK:
Would be better to tweak Fortress to block `'BlazingPinions'` damage type rather than change the mechanics of Blazing Pinions.

pledbrook:
For the moment, I've added 'BlazingPinions' to the list of Fortress damage immunities.

BlackRose39:
Very well, in that case, I'll close this.

PR #621, opened 2019-09-27
#576 Fixed issue where Locked On bonus was being lost

### LINKED ISSUE #576: Locked On Bug
Seems like having another xcom member take an action in between the two shots cancels Locked On bonus. In the attached save, Frosty is Locked On to the viper due to an OW attack. If you take 2 shoots immediately, both get the locked on bonus, but if you have any other xcom member attack, OR trigger the muton's overwatch, she loses the bonus. Happens if you take the shot before her first shot, or her second shot, so Light Em Up does not seem to be the cause.

[save_Locked On Bug.zip](https://github.com/long-war-2/lwotc/files/3637205/save_Locked.On.Bug.zip)

Favid:
Fixed issue where Locked On bonus was being lost if another shot was taken in between the Locked On soldier's shots. Closes #576


pledbrook:
Thanks for this one!

PR #624, opened 2019-09-28
#505 Added an Advanced Option to allow XCOM soldiers to purchase multiple perks per rank

### LINKED ISSUE #505: Suggestion: Make an SW option so you can buy all soldiers abilities with AP (like WOTC)
Hey, I know there will be balance involved (which I think should not be accommodated for this option as it will not be the default one), but one of the things that was really fun on WOTC was that the fact that the more you would invest in them, you could see the results with improving their effectiveness. This was balanced out by the fact that AP is a shared and balanced element so you can't make everyone into killing machines.
I'm pretty sure this will be one of the first things people will mod out the behaviour, so why not integrate it as a option so they can choose what experience they want? Of course if it's not that much trouble and would not affect time into other things.
Kind Regards

As BlackRose mentioned : For a temporarily workaround, I find that you can head to xcompromotionuimod.ini and set the "abilitiesperrank" to 0 (currently only for individual classes), and it should work exactly as it should that the original promotion mod does where you can pick as many abilities per rank as much as you like. For modded classes, you'll have to add the class manually. Its recommended you increase the ap on promotions too, or else you dont get even ap for one perk, to balance it out (because of the costs).

Favid:
Added an Advanced Option to allow XCOM soldiers to purchase multiple perks per rank, as long as the Training Center has been built. Closes #505

### DIFF
--- a/LongWarOfTheChosen/Localization/BetterSecondWaveSupport.int
+++ b/LongWarOfTheChosen/Localization/BetterSecondWaveSupport.int
@@ -1,11 +1,11 @@
 ; LWOTC Needs Translation
 [EnableChosen SecondWaveOptionObject]
 Tooltip="Enables the Chosen, who will activate after you complete the first network tower mission."
-Description="The Chosen: Enables the Chosen"
+Description="The Chosen: Enables the Chosen."
 ; End Translation
 
 ; LWOTC Needs Translation
 [AllowSameRankAbilities SecondWaveOptionObject]
-Tooltip="Allows XCOM soldiers to purchase multiple perks of the same rank with AP"
-Description="Allow Same Rank Abilities: XCOM soldiers can purchase multiple perks of the same rank with AP once the Training Center has been built."
+Tooltip="Allows XCOM soldiers to purchase multiple class abilities of the same rank with AP."
+Description="Allow Same Rank Abilities: Allows XCOM soldiers to purchase multiple class abilities of the same rank with AP once the Training Center has been built."
 ; End Translation
--- a/LongWarOfTheChosen/Localization/BetterSecondWaveSupport.int
+++ b/LongWarOfTheChosen/Localization/BetterSecondWaveSupport.int
@@ -1,3 +1,11 @@
+; LWOTC Needs Translation
 [EnableChosen SecondWaveOptionObject]
-Tooltip="Enables the Chosen, who will activate after you complete the first network tower mission"
+Tooltip="Enables the Chosen, who will activate after you complete the first network tower mission."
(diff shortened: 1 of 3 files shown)


LongWarOfTheChosen/Localization/BetterSecondWaveSupport.int
```diff
@@ -1,3 +1,11 @@
+; LWOTC Needs Translation
```
pledbrook: Oops. Missed this.

LongWarOfTheChosen/Localization/BetterSecondWaveSupport.int
```diff
@@ -1,3 +1,11 @@
...
 [EnableChosen SecondWaveOptionObject]
-Tooltip="Enables the Chosen, who will activate after you complete the first network tower mission"
+Tooltip="Enables the Chosen, who will activate after you complete the first network tower mission."
 Description="The Chosen: Enables the Chosen"
+; End Translation
+
+; LWOTC Needs Translation
```
pledbrook: I would honestly treat them as a single translation block right now.

LongWarOfTheChosen/Localization/BetterSecondWaveSupport.int
```diff
@@ -1,3 +1,11 @@
...
+[AllowSameRankAbilities SecondWaveOptionObject]
+Tooltip="Allows XCOM soldiers to purchase multiple perks of the same rank with AP"
```
pledbrook: OK, so why did my tooptip require a full stop, but your doesn't 😛

LongWarOfTheChosen/Localization/BetterSecondWaveSupport.int
```diff
@@ -1,3 +1,11 @@
...
+Description="Allow Same Rank Abilities: XCOM soldiers can purchase multiple perks of the same rank with AP once the Training Center has been built."
```
pledbrook: I would clarify that this applies to _class_ abilities/skills/perks. Not quite sure which of those words to standardise on.

LongWarOfTheChosen/Localization/BetterSecondWaveSupport.int
Favid: Oops, all tooltips and descriptions should have punctuation to match the vanilla ones.

LongWarOfTheChosen/Localization/BetterSecondWaveSupport.int
Favid: Good catch, I think abilities would be best here.

PR #631, opened 2019-09-28
Nerf Chain Lightning

Grobobobo:
As discussed, chain lightning will have -50 internal aim for the ability. Note the Electroshock ability, which disorients the target even on miss and the Stun Gunner ability, which gives +20-+30 to hit. I've been told the nerf is insufficient, so might change the AP cost to 2 as well.


pledbrook:
I would consider an aim penalty *or* a 2 AP cost for a nerf, but I think both might just be unusable unless you have Bunker Buster. Which may be the point. If you were to change to 2AP, I would drop the aim penalty some. Honestly don't know.

Grobobobo:
Just remember that it still can be used from stealth.

pledbrook:
Added in [this commit](https://github.com/long-war-2/lwotc/commit/88e2bc7bc35779ff9105fddcc3bc7d31f4edb876). Thanks!

PR #633, opened 2019-09-28
Give Sectopods Resilience

Grobobobo:
As Discussed. I think.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/LWTemplateMods.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/LWTemplateMods.uc
@@ -1945,6 +1945,9 @@ function GeneralCharacterMod(X2CharacterTemplate Template, int Difficulty)
 			Template.Abilities.AddItem('CoupdeGrace2');
 			Template.Abilities.AddItem('Whirlwind2');
 			break;
+		case 'Sectopod'
+			Template.Abilities.AddItem('Resilience');
+
 		// Should turn off tick damage every action
 		case 'ViperKing':
 		case 'BerserkerQueen':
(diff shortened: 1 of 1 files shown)


BlackRose39:
I agree, and also, give them the "easytohit," since they're big targets.

Grobobobo:
That will depend on PaledBrook.

ansabalirai:
As commented on github, I would suggest linking it to a DE instead and applied to a configurable percentage. This would also make it possible to extend it to other massive enemies, e.g. cyber ball.

Grobobobo:
The point is to make them more resistant to Kubikuri, so making it a DE doesn't make much sense.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/LWTemplateMods.uc
```diff
@@ -1945,6 +1945,9 @@ function GeneralCharacterMod(X2CharacterTemplate Template, int Difficulty)
 			Template.Abilities.AddItem('CoupdeGrace2');
 			Template.Abilities.AddItem('Whirlwind2');
 			break;
+		case 'Sectopod'
+			Template.Abilities.AddItem('Resilience');
```
pledbrook: You're missing a `break;` here.

pledbrook:
I'm tempted to give all robotics a slightly weaker Resilience, just to make them even more distinctive. That would be my ultimate goal: make enemy units as distinctive as possible, with significant strengths and significant weaknesses. Ideally with multiple ways to deal with them.

Grobobobo:
I fixed that in the second commit. I guess that's why they should be informative, huh?

pledbrook:
I would be OK with an easy-to-hit mod, or a negative defence. Not sure about giving them 'CivilianEasyToHit' though :smile:

Grobobobo:
I don't even think civilians should have 'CivilianEasyToHit'.

ansabalirai:
Well robots already do not have the normal 40% flank percentage that you get on other aliens,, so... bluescreen strat incoming?

pledbrook:
I think the current Resilience may be too strong, and it's worth bearing in mind that the Stocks will be nerfed too, so stacking crit is a bit harder.

Regardless, there are still several anti-robotic options, lots of shred (although coilgun shred needs fixing), plus several sources of hack defence reduction with shutdown, etc. Let's see how it plays out.

PR #636, opened 2019-09-28
Make supply raids require 100% infil

Grobobobo:
Well, make ALL corpse missions that aren't invasions require 100% infil, As compensation, some of these missions have better detection rates. Seems experimental, but worth a try.


BlackRose39:
My only concern is that what happens if you don't have enough time to infiltrate to 100%? Will the mission be unlaunchable?

Grobobobo:
It should give you an option to either boost infiltration, or abort. Should is the key word here though.

pledbrook:
I'll want to test this before merging.

LongWarOfTheChosen/Config/XComLW_InfiltrationSettings.ini
```diff
@@ -60,6 +60,14 @@ RequiredInfiltrationToLaunch=-0.0000001f
 +MissionsRequiring100Infiltration=AdventFacilityFORGE_LW
 +MissionsRequiring100Infiltration=AdventFacilityPSIGATE_LW
 +MissionsRequiring100Infiltration=CentralNetworkBroadcast_LW
++MissionsRequiring100Infiltration=SupplyLineRaidATT_LW
++MissionsRequiring100Infiltration=SupplyLineRaidTrain_LW
++MissionsRequiring100Infiltration=SupplyLineRaidConvoy_LW
++MissionsRequiring100Infiltration=TroopManeuvers_LW
++MissionsRequiring100Infiltration=SecureUFO_LW
```
pledbrook: I'd like to limit this to just Supply Raids for now.

Grobobobo:
Definitely.

PR #637, opened 2019-09-28
Small corrects for new weapon tier soundrange and invdamage

nightmaRok:


LongWarOfTheChosen/Config/XComGameData_WeaponData.ini
```diff
@@ -1957,7 +1957,7 @@ VEKTOR_MAGNETIC_RANGE[23] = 0
 VEKTOR_MAGNETIC_RANGE[24] = 0
 VEKTOR_MAGNETIC_RANGE[25] = 0
 
-+VEKTOR_COIL_RANGE[0]=0
++VEKTOR_COIL_RANGE[0]=30
```
pledbrook: Why is this 30 rather than 25?

nightmaRok:
VEKTOR_BEAM_RANGE[0] = 30
VEKTOR_MAGNETIC_RANGE[0] = 30

So why not

nightmaRok:
Also i look now at files, coil ranges i think might be like beam

pledbrook:
Well, I don't understand what the issue is with 0 aim bonus at 0 range, since you can't shoot at 0 range. So why this change in the first place? And I asked about 30, because the only time I've seen a positive value for the 0-range entry, it's been the same as the one at 1-tile range.

nightmaRok:
I see this setting for lw2 weapons in this file, if you think that dont needed i clear it
SHORT_CONVENTIONAL_RANGE[0] = 60
MEDIUM_CONVENTIONAL_RANGE[0] = 30
LONG_CONVENTIONAL_RANGE[0] = -30

All weapon types have this, even for aliens
FLAT_CONVENTIONAL_RANGE[0] = 30

Grobobobo:
The [0] is apparently for some edge cases with elevation.

pledbrook:
@nightmaRok So in all those examples you quote, the `[0]` entry has the same value as `[1]`. I think that's intentional. That's why I suspect, but I'm not sure, it should be 25 for Vektors.

MrNiceUK:
I suspect with the integer truncation of the float distances, they are just protecting against the distance rounding down to zero in edge cases, better safe than sorry!

pledbrook:
Added in [this commit](https://github.com/long-war-2/lwotc/commit/3f0e3fb98502df8c996fb434474ed23f2fdb67fc). Thanks!

PR #639, opened 2019-09-29
Rebalance XCOM (random) abilities

pledbrook:
Soldier classes now have a choice of 3 T1, 3 T2, and 1 T3 ability, changed from 2 T1, 3 T2, and 2 T3. The costs of T1/T2/T3 have also changed to 10/20/40.

This means standard and above average combat intelligence can't cover a T3 ability alone. It also favours a bit more distinction between soldiers at the T1 level, because those are the ones most soldiers will be able to afford. And now it's a trade off of one T2 ability vs 2 T1 abilities.

Hopefully this will mean more disintictive, but not over-powered soldiers.


pledbrook:
Added in [this commit](https://github.com/long-war-2/lwotc/commit/9150efdf5b7876359e0ef7467e1b179d76b50d6d).

PR #647, opened 2019-10-02
Localization for Templar Stun strike

nightmaRok:


nightmaRok:
Martox give me this changes, so now its look like this

pledbrook:
Thanks!

PR #672, opened 2019-10-07
Fix the messed up config names

Grobobobo:
For some reason, I screwed up the MEDIUM_LASER_RANGE with SPARKRIFLE_LASER_RANGE.

Making the laser variants have no range tables, same with coils.
 
This should fix that.


pledbrook:
Fixed in [this commit](https://github.com/long-war-2/lwotc/commit/ccdf7f4a26357142d5c3aa24de34617cf2c54fbb). Thanks for the code.

PR #680, opened 2019-10-13
#471 Updated new abilities to interact properly with lost, chosen, an…

### LINKED ISSUE #471: Technical Rocket Launcher does not summon lost like other explosives
On a mission where lost spawn, the Technical's rocket launcher should summon lost like other explosives (e.g. grenades) do. This is a simple thing to fix, just need to add a few lines to the ability template. The LW2 Secondary Weapons mod does this correctly so just look at that for reference. Also, need to fix any other abilities and item that have this same problem.

Favid:
…d shadow

Updated new ability templates to have proper values for SuperConcealmentLoss, ChosenActivationIncreasePerUse, and LostSpawnIncreasePerUse. Closes #471


pledbrook:
Thanks so much for doing this! Out of interest, does this fix the issue with Precision Shot only having a 10% chance to break a Reaper's Concealment the first time?

Favid:
> Thanks so much for doing this! Out of interest, does this fix the issue with Precision Shot only having a 10% chance to break a Reaper's Concealment the first time?

I didn't test Precision Shot specifically, but yeah this should fix that.

Issue #685, opened 2019-10-14
Additional Faction Soldiers


TheReigningKing:
Ability to recruit extra (2 total) faction soldiers from each faction, even if they are not your starting faction.

This should require a MSGT soldier to accomplish.

See RealityMachina mod: https://steamcommunity.com/sharedfiles/filedetails/?id=1126788564

Fallguy55:
I don't think it should require master sergeant. Maybe gunnery sergeant? Or maybe it should require a mid rank officer to complete the mission. I feel as if two of each faction soldiers isn't so overpowered as it would have been in vanilla, being as though your forces are spread much much wider in the long war setting.

PR #727, opened 2019-11-06
#685 Allow 2 hero soldiers per faction

Grobobobo:
Allows 2 hero soldiers per faction,
Uses the code from Realitymachina's mod
No change to the rank requiriment (for now)


LongWarOfTheChosen/Src/LW_FactionBalance/Classes/X2DownloadableContentInfo_LW_FactionBalance.uc
```diff
@@ -28,6 +28,8 @@ static event OnPostTemplatesCreated()
 	UpdateDeflect();
 	UpdateShadow();
 	UpdateRemoteStart();
+	//Allows the 2-2-2 faction soldier roster
+	EditReward();
```
pledbrook: This doesn't really make it obvious what's being changed. Edit what reward? What's 2-2-2 mean? (Yes, I know, but only because I've seen you reference it in Discord). I would just call the function `AllowTwoSoldiersFromEachFaction()`. Yes it's more wordy, but it's pretty clear. You don't have to go with that specific function name, but please use something that's a bit more informative. (The other function names might benefit from similar treatment, but let's leave those alone for now).

LongWarOfTheChosen/Src/LW_FactionBalance/Classes/X2DownloadableContentInfo_LW_FactionBalance.uc
```diff
@@ -346,6 +348,52 @@ static function EditParry(X2AbilityTemplate Template)
...
+	if (FactionState != none)
+	{
+
+
+		NumFactionSoldiers = FactionState.GetNumFactionSoldiers(NewGameState);
+		foreach History.IterateByClassType(class'XComGameState_CovertAction', ActionState)
+		{
+			if(ActionState.GetMyTemplateName() == 'CovertAction_RecruitExtraFactionSoldier' && (ActionState.GetFaction().GetReference().ObjectID == FactionState.GetReference().ObjectID) && ActionState.bStarted && !ActionState.bCompleted) //this is dumb but we have to account for this
+				NumFactionSoldiers += 1;
+		}
```
pledbrook: I don't understand why this code is necessary. It seems like a sledgehammer solution. Is there a bug in the base game it's trying to get round?

Grobobobo:
I should ask realitimachina, but I probably know why - It's because you can have multiple "recruit faction soldiers" CA at the same time, so it's effectively making sure you can have only CA reward at the same time.

LongWarOfTheChosen/Src/LW_FactionBalance/Classes/X2DownloadableContentInfo_LW_FactionBalance.uc
pledbrook: This shouldn't be needed. The community highlander has a fix for duplicate covert actions appearing while one is already running. So I'd prefer to remove this hack on the assumption it's not needed. If we need to fix it later on, so be it.

Grobobobo:
I've made qute a mess in this PR, I'll try readding it.

Issue #699, opened 2019-10-22
Advent Vanguard Close Combat Specialist don't work


nightmaRok:
[save_Advent_CCS_1.zip](https://github.com/long-war-2/lwotc/files/3756062/save_Advent_CCS_1.zip)

![001](https://user-images.githubusercontent.com/51829747/67292972-6d9d5f00-f4ec-11e9-892f-6a1052f27623.JPG)

BlackRose39:
Just to add, the elite sidewinder doesn't move after it triggers "Hit and sliter."

Grobobobo:
Almost none of the "active" alien abilities worked. Likely an AI issue.

BlackRose39:
CCS needs to apply to the primary weapon slot of the unit as well as other abilities that require a slot.

Grobobobo:
That's it? That's what wasn't working? I'm 99% sure that's a config change

pledbrook:
See [this pull request](https://github.com/long-war-2/lwotc/pull/739).

PR #739, opened 2019-11-13
#699 Fixed Alien CCS

Grobobobo:
Fixed Aliens not Triggering CCS by tying
the ability to their primary weapon.
Needed to slightly change the apply
funtion for the DE version to work, and
I'm not sure I've changed it in the most efficient way.


Grobobobo:
I forgot to mention, but I've taken the liberty to increase the vanguard's clip size because LW2 CCS takes 2 ammo per shot.

pledbrook:
I'll need some time to look at this. It's not clear what's happening at all and I'm not familiar with the LW2 code in any case.

Grobobobo:
Basically, the reason why both variants of CCS weren't working is because they weren't tied into any slot.  In vanguard's case, assigning the ability to it's primary weapon template is enough, but the dark event one requires you to assign it to the weapon during the tactical begin phase (since that's when the DE ability bonuses are assigned). The ApplyToPrimaryWeapon bool did nothing, so i changed it to specify assigning abilities to unit's primary weapons. Not sure if that helped clarify things.

Grobobobo:
Closing to reupload later.

PR #744, opened 2019-11-15
Additional Attachment changes

Grobobobo:
Regarding the changes we discussed on discord.
That includes:
- Makes auto loaders and ex mags
mutually exclusive
- Updates the values for laser sights,
 which weren't in previously
- Replaces stock's effect with
a weaker grazing fire
- grants an aim-only steady weapon
to all sniper rifles
- decreases the cost of some of the
 attachment proving ground projetcs
(namely attachments 3 ->1 and slight
cost reducition for scopes

These are slightly experimental, and
will need to be tested from a balancing
standpoint.


pledbrook:
The change to the Laser Sight chances is news to me. I deliberately left them as they were as there are fewer opportunities to stack crit compared to aim. I don't have a strong opinion either way. What was the reasoning? Simply consistency with the aim attachments?

Grobobobo:
Mainly the consistency, but also the fact that they get like double the bonus at point blank. +30 crit seems a bit strong all things considered.

pledbrook:
It's +10 bonus at close range. We could just remove the range bonus. I don't really understand why it's there to be honest.

Grobobobo:
We can remove both :). More seriously, I think a laser sight should be a viable-ish attachment on all weapons, (It's only problematic on SR because of kubikiri), So I'm much more inclined towards removing the close range bonus. Should I make a commit here?

pledbrook:
Yes, you can add the change to this pull request.

PR #757, opened 2019-11-23
Updated XComSchedules.ini for Faceless spawning change

Anthogator:
Currently reverted to faceless only spawning on Neutralize missions unless the faceless DE is active. Had to remove spawns from many GOp schedules to accomplish this.

Apologies for github not showing the exact changes - I am still learning how to use github in a friendly manner!


pledbrook:
I'm sorry. I can't merge this. I suspect the problem is that the line endings in GitHub are different from the ones you uploaded, so it considers *all* the lines have changed.

If you can, try to upload the file again with Unix line endings rather than Windows ones. Any half-decent text editor (i.e. not Notepad) will allow you to do that.

Anthogator:
I understand. I am on vacation for the next while but will attempt to get a better, mergeable pull request for you when I can.

PR #779, opened 2019-11-28
Fixed Alien CCS

Grobobobo:
Partly #699 but also for the Dark Event Close Combat specialist. The pull request is reuploaded.


pledbrook:
I've reimplemented this for better compatibility with other mods and so that it also works for other abilities that require a primary weapon.

PR #787, opened 2019-11-30
LW_Overhaul.ita fully localized

SilentSlave:


pledbrook:
Does this supercede the other pull request? If so, could you please close the other one.

PR #788, opened 2019-11-30
ITA Localization

SilentSlave:
This should be the finished version? I can understand it may seem chaotic, but it does everything right in the end.


LongWarOfTheChosen/Localization/LW_Overhaul.ita
```diff
@@ -1,196 +1,202 @@
+;LWotC Translated
```
pledbrook: Sorry, you only need to do this where the `.int` file has a `; LWOTC Needs Translation` block. If you have made other changes to fix existing translations, don't bother marking those.

LongWarOfTheChosen/Localization/LW_Overhaul.ita
```diff
@@ -1,196 +1,202 @@
...
+strIncompatibleModsTitle="MOD DUPLICATE PRESENTI"
+strIncompatibleModsText="Rilevate mod duplicate:\n%modnames%\nQueste mod sono già state integrate in questa mod e dovrebbero essere disabilitate dal launcher. Caricandole potrebbero verificarsi effetti negativi, come scarse prestazioni o instabilità. Procedere a proprio rischio."
 
 [UIMission_LWCustomMission]
 m_strMissionDifficulty_start="ATTIVITÀ NEMICA BASE"
 
 [ProtectRegionEarly X2LWAlienActivityTemplate]
 ActivityName="ProtectRegion1Activity"
-ActivityObjectives[0]="Trova un indizio"
+ActivityObjectives[0]="Inizia a liberare la regione"
```
pledbrook: So this is the line that needs to be warpped with
```; LWOTC Translated
...
; End Translated
```
because it's the one in _LW\_Overhaul.int_ that has `; LWOTC Needs Translation`. Does that make sense?

pledbrook:
Merge as [this commit](https://github.com/long-war-2/lwotc/commit/685518f1723d1b043a552e54f88b34caafd7a4e7)

PR #840, opened 2020-01-06
Fix the incorrect sound range config names

Grobobobo:
Some of the weapons have incorrect sound range config names, likely leading to them not alerting any pods in the game. This change should fix that.


pledbrook:
Is this change necessary? Didn't think UnrealScript was case sensitive.

Grobobobo:
I found it because the compiler displayed a warning about it, so probably

Issue #805, opened 2019-12-09
Upcoming Chosen Changes Part 2 - Standard mission behaviour and abilities


Grobobobo:
This issue is about implementation of [my chosen changes proposal](https://drive.google.com/file/d/1wRUuNnb8g0eGpPsFKZalZWqaIrGv74IT/view) and less the discussion about it, as I feel it has been talked about to death already.

I'd say that this part is the most complicated, but luckily a lot of it has already been done by other modders, so there's that.

Stat changes are easy, no talking nescessary.

As stated previously, I intend to give warlock abilities from A better Chosen mod by DerBK. There are even already done prime reactions there!

Moreover, I'll be adding two abilities to Hunter:

Sniper rifle TranqShot that he'll use as a replacement for tracking shot when unactivated.

The ability that let's him teleport to a dazed enemy and use extract knowledge on it to capitalize on the ability above.

There are two major challanges here: The Heavy Daze Effect, and the AI changes. The Heavy daze could be simply made as a seperate effect, but I'm worried how the soldier that's both dazed and heavy dazed will act.

As for AI, well, that'll largely depend, I'll likely have to test the shit out of it.

HarmlessOrange:
I think Retaliation should be the Chosen attacking the Haven's residents, injuring or even killing some of your guys in that Haven. It could be through RNG or a special rendezvous-like mission, but with 1 pod and a Chosen instead Faceless. (And if proves to be too hard, you could downgrade his stats for these mission. Like they're holding back because they're not XCOM."

If by RNG then have the Advisor affect the outcome. A scientist would save the affected, making them injured instead killed. Engineer would lower the chance of his Haven to be hit. A Soldier would add a chance for the retaliation to fail, but the chances would be affected by his Rank. So you have more incentive to plant your higher-ranking troops into Havens.

Keep in mind that the Chosen might kill the Advisor as well.

...And I have no idea if all of this is even doable.

If possible then Heavy Daze could simply remove normal Daze before applying itself. So if Soldier is Dazed and then Heavy Dazed, Normal Daze is healed and Heavy Daze goes into its place.

As for weapons... I will need some time to think of something.

Anxoczt:
I rather not have Chosen facing of against resistance members if said members are absolute shit when compared to XCOM troops that die if someone looks in their general direction. I would prefer to keep Chosen combat against XCOM troops only.

Grobobobo:
Scrapped.

Issue #804, opened 2019-12-09
Upcoming Chosen Changes Part 1 - The strategic Layer


Grobobobo:
This issue is about implementation of [my chosen changes proposal](https://drive.google.com/file/d/1wRUuNnb8g0eGpPsFKZalZWqaIrGv74IT/view) and less the discussion about it, as I feel it has been talked about to death already.

Overall Strategy layer changes seem like the easiest to implement overall, Making the Chosen level up with force level rather than training means changing the training activity template, and adding a function to `ScheduledReinforcemetsCompleted` that levels up chosen as the Force level reaches certain thresholds. 

The rest of activities are very similiar, They have a template, I change the template, done.

Mixing up their activities is just mostly setting up their priorities and which action is available is just messing with priorities and which activity can be done at which knowledge level.

I want Training to be from level 1 (would generate a nice negative feedback loop, if you constantly kick their asses and they stay at low knowledge level they decide to git gud rather than to disrupt you) Retribution to be from level 2, Dark Event from level 3, Sabotage from level 4, Avenger assault for max.

Overall by far the easiest part of the overhaul, I don't think there's much to talk here.

Grobobobo:
Scrapped.

PR #842, opened 2020-01-10
Merge the Experimental build changes from the fork 

Grobobobo:
In case you'd want it. 
contains mostly issues #804 and #805. 
I didn't create the PR because some of the effect removing still uses indexing, and I'm sure I missed an ident somewhere.


Grobobobo:
Looks like the PR was not needed.

PR #853, opened 2020-01-28
Remove the guaranteed faceless from a starting region and make rebels have an intel job by default

Grobobobo:
Remove the guaranteed faceless from a starting region and make rebels have an intel job by default.


pledbrook:
This is open because it's currently merged into `experimental` for play testing. Will close if and when it it's merged into `master` (should be as it seems a reasonable change).

pledbrook:
Merged with [this commit](https://github.com/long-war-2/lwotc/commit/8863de02d897df6c8ce2edce81180c664bc974a2)

PR #877, opened 2020-02-14
Update install wiki link in README

arrayjam:
Now you can open the install page and still have a spare action left to hunker down.


pledbrook:
Thanks! Sorry for the delay in merging.

PR #906, opened 2020-03-17
Change the way steady weapon works

Grobobobo:
Kubikiri's gone, so it can be stronger now.


Grobobobo:
Let's not do this change for the next version after all.

PR #916, opened 2020-03-25
Improve The design of steady weapon

Grobobobo:
- Steady weapon is now +20/+20 at all weapon tiers, but assault rifles, cannons and spark rifles get 1.5 times the bonus and shotguns get 2.0
- Steady Weapon no longer loses its effect when the soldier is wounded
- Steady Weapon is now automatically granted to rifles, cannons, shotguns, sniper rifles and spark rifles
- Steady Weapon can no longer be used in concealment

These changes should encourage more agressive planning for the next turn, making it viable on more weapon types (and lower aim soldiers, like rookies) While also decreasing the cheesy aspects of the skill.


LongWarOfTheChosen/Src/LW_PerkPack_Integrated/Classes/X2Effect_SteadyWeapon.uc
```diff
@@ -105,18 +110,49 @@ static function EventListenerReturn SteadyWeaponWoundListener(Object EventData,
 function GetToHitModifiers(XComGameState_Effect EffectState, XComGameState_Unit Attacker, XComGameState_Unit Target, XComGameState_Ability AbilityState, class<X2AbilityToHitCalc> ToHitType, bool bMelee, bool bFlanking, bool bIndirectFire, out array<ShotModifierInfo> ShotModifiers)
 {
 	local ShotModifierInfo ShotInfo;
+	local XComGameState_Item SourceWeaponState;
+
+	SourceWeaponState = AbilityState.GetSourceWeapon();
 
 	if(!bMelee && AbilityState.SourceWeapon == EffectState.ApplyEffectParameters.ItemStateObjectRef)
```
pledbrook: I would put a space after `if` here too, along with the new `if` statements below.

LongWarOfTheChosen/Src/LW_PerkPack_Integrated/Classes/X2Effect_SteadyWeapon.uc
```diff
@@ -105,18 +110,49 @@ static function EventListenerReturn SteadyWeaponWoundListener(Object EventData,
...
+		if(default.WEAPON_CATEGORIES_HIGH_BONUS.Find(SourceWeaponState.GetWeaponCategory()) != INDEX_NONE)
+		{
+			ShotInfo.ModType = eHit_Success;
+			ShotInfo.Reason = FriendlyName;
+			ShotInfo.Value = Aim_Bonus * default.HIGH_BONUS_MODIFIER;
+			ShotModifiers.AddItem(ShotInfo);
+		
+			ShotInfo.ModType = eHit_Crit;
+			ShotInfo.Reason = FriendlyName;
+			ShotInfo.Value = Crit_Bonus* default.HIGH_BONUS_MODIFIER;
```
pledbrook: Why not put the aim and crit values in the `if` blocks and the rest after them? There's a lot of code duplication here since the `ModType`s, `Reason`s and `AddItem()`s are the same for all three branches.

Grobobobo: I did it because Shotinfo is one variable, and it would just override each other. But yeah, I think I'm going to give crit a seperate variable.

Grobobobo:
Adressed the issues in the review.

PR #921, opened 2020-03-28
Miscellaneous bug fixes

pledbrook:
Contains fixes for enemies shooting rockets/spitting poison through walls and solves the various issues with Revival Protocol granting extra action points to the target when it shouldn't.

I have made the Micro Missiles' AoE profiles require line of sight along with the other AoE abilities, but I'm not sure whether this will prevent MEC Archers from using their micro missiles at squadsight range, which is kind of the point of the unit. Would like to just include the fix, though, and see whether it has any noticeable effect on the Archers.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_PerkPack_Integrated/Classes/X2Condition_RevivalProtocol_LW.uc
+++ b/LongWarOfTheChosen/Src/LW_PerkPack_Integrated/Classes/X2Condition_RevivalProtocol_LW.uc
@@ -1,0 +1,22 @@
+//---------------------------------------------------------------------------------------
+//  FILE:    X2Condition_RevivalProtocol_LW.uc
+//  AUTHOR:  Peter ledbrook
+//  PURPOSE: Allows Revival Protocol to also target stunned units.
+//---------------------------------------------------------------------------------------
+class X2Condition_RevivalProtocol_LW extends X2Condition_RevivalProtocol;
+
+event name CallMeetsCondition(XComGameState_BaseObject kTarget)
+{
+	local XComGameState_Unit TargetUnit;
+    local name Result;
+
+	TargetUnit = XComGameState_Unit(kTarget);
+
+    Result = super.CallMeetsCondition(kTarget);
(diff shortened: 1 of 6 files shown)


LongWarOfTheChosen/Src/LW_PerkPack_Integrated/Classes/X2Condition_RevivalProtocol_LW.uc
```diff
@@ -0,0 +1,22 @@
...
+
+event name CallMeetsCondition(XComGameState_BaseObject kTarget)
+{
+	local XComGameState_Unit TargetUnit;
+    local name Result;
+
+	TargetUnit = XComGameState_Unit(kTarget);
+
+    Result = super.CallMeetsCondition(kTarget);
+    if (Result == 'AA_UnitIsNotImpaired' && TargetUnit != none && TargetUnit.IsStunned())
```
Grobobobo: It might be better at this point to put both effect conditions and effects to be removed in config arrays (and then check if unit is affected by the effect name in the array), in case we want to change which effects it works on (for example I wanted to actually remove daze from it for a while), or just to make it easier to make it work with other mod effects (hypothetically something like heavy daze)

pledbrook:
The `bRequireLoS` parameter definitely works for poison spit. I tested it pretty thoroughly.

pledbrook:
I also tried out another change that LW2 introduced, which was to allow squad visibility in the line-of-sight check, but then if a squad member has line of sight, then Vipers can start spitting through walls again. So if there's an issue with MEC Archers, we'll have to work out how to give them Squadsight without breaking them.

LongWarOfTheChosen/Src/LW_PerkPack_Integrated/Classes/X2Condition_RevivalProtocol_LW.uc
pledbrook: Eh. Thought about it, but it's more effort than I'm willing to invest right now. If a configurable array is warranted at a later date, we can do it. But you also have to deal with the `X2Effect_RemoveEffects` array.

PR #926, opened 2020-04-02
Dark events as sit reps

pledbrook:
Turns some of the dark events into sit reps that simply have an independent chance to apply to a mission while the dark event is active. This also makes the dark event sit reps part of the general sit rep pool, meaning they can apply to missions even when the dark event isn't active.

Also made a couple of other changes related to sit reps (but not dark events).


LongWarOfTheChosen/Config/XComGameData.ini
```diff
@@ -1295,20 +1295,24 @@ MedicalModeHealRateScalar[3]=0.25
...
 -SitReps="PrivateChannelsSitRep"
 -SitReps="MentalFortitude"
--SitReps="DarkEventBarrierSitRep"
--SitReps="DarkEventDarkTowerSitRep"
+; -SitReps="DarkEventBarrierSitRep"
+; -SitReps="DarkEventDarkTowerSitRep"
 ;-SitReps="ChallengeNoReinforcements" DON'T Disable this as it only applies to TQL and skirmish mode and swarms you with redscreens. Really annoying
 
-
-+SitReps=LowVisibility
```
Grobobobo, line 1303: While we're at it, can we remove location scout? That sitrep is incredibly cheesy

LongWarOfTheChosen/Config/XComGameData.ini
```diff
@@ -1317,6 +1321,84 @@ DisplayEffects="LethargyEffect"
...
+MinimumForceLevel=6
+bNegativeEffect=true
+
+NegativeEffects="DarkEventVigilanceEffect"
+bExcludeFromStrategy=false
+
+[TheLost X2SitRepTemplate]
++NegativeEffects="TheLostSitRepEffect"
+
+; Keep the Lost off of untimed missions
```
Grobobobo, line 1390: Should add:  SupplyLineRaid_LW and Sabotage_LW

pledbrook: Wait, what? They're both untimed, right? So they shouldn't be in the list (although the Facility mission does have a soft timer with cavalry reinforcements).

LongWarOfTheChosen/Config/XComGameData.ini
pledbrook: Yeah. I was wondering about a sit rep that just gives all your soldiers Tracking (say it's something like motion tracking from somewhere). Might be a suitable replacement.

LongWarOfTheChosen/Config/XComGameData.ini
Grobobobo: ...Nevermind, got it confused with MISSION_TYPE_SIT_REP_EXCLUSIONS

pledbrook: Also, if you think changes need to be made, as in this case, don't hesitate to pick the Request Changes options when submitting the review.

LongWarOfTheChosen/Config/XComGameData.ini
pledbrook: I'll remove it once we have a replacement. It makes a nice change of pace for the moment.

PR #931, opened 2020-04-05
Rebalance failure risk chances for CAs

pledbrook:
Early covert actions should be slightly more likely to succeed, but the risk chance ramps up with force level more quickly. This is to prevent mid-game covert actions being reliably done with just LCPLs and CPLs. The effect of each rank has also been reduced for the same reason.

Lastly, I've added a cap on the FL increases so that MSGT squads can reliably complete covert actions throughout the game, rather than then having significant chances to fail by FL20.

### DIFF
--- a/LongWarOfTheChosen/Config/XComLW_Overhaul.ini
+++ b/LongWarOfTheChosen/Config/XComLW_Overhaul.ini
@@ -1581,15 +1581,18 @@ BLACK_MARKET_PERSONNEL_INFLATION_PER_FORCE_LEVEL=0.075
 BLACK_MARKET_SOLDIER_DISCOUNT=0.4
 
 [LW_Overhaul.X2EventListener_Headquarters]
-CA_RISK_REDUCTION_PER_RANK[0]=15   ; Rookie
-CA_RISK_REDUCTION_PER_RANK[1]=12   ; Veteran
-CA_RISK_REDUCTION_PER_RANK[2]=10   ; Commander
-CA_RISK_REDUCTION_PER_RANK[3]=10   ; Legend
+CA_RISK_REDUCTION_PER_RANK[0]=12   ; Rookie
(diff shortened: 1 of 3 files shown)


LongWarOfTheChosen/Config/XComLW_Overhaul.ini
```diff
@@ -1581,15 +1581,18 @@ BLACK_MARKET_PERSONNEL_INFLATION_PER_FORCE_LEVEL=0.075
...
+CA_RISK_REDUCTION_PER_RANK[2]=8    ; Commander
+CA_RISK_REDUCTION_PER_RANK[3]=8    ; Legend
+
+CA_RISK_INCREASE_PER_FL[0]=4       ; Rookie
+CA_RISK_INCREASE_PER_FL[1]=4       ; Veteran
+CA_RISK_INCREASE_PER_FL[2]=4       ; Commander
+CA_RISK_INCREASE_PER_FL[3]=4       ; Legend
+
+; The FL at which the failure risk no longer increases
+CA_RISK_FL_CAP=12
```
Grobobobo, line 1595: That seems a little low, especially since risk reduction happens per person. I'd make it 14, maybe even 16.

pledbrook: Higher than 12 means that a full squad of MSGTs will not have 100% success chance on the hardest covert actions. It's a base chance of failure of 80% + FL12 (4 * 12) = 128%.

Of course, having a small chance of failure for MSGTs on the hardest covert actions isn't necessarily a major problem. I don't really have a strong opinion on this. Could up it to FL16, which would mean a maximum failure chance of 16%. 🤷‍♂

Grobobobo: Eh, let's try it as is for now, pretty easy to change if it turns out to be boring.

PR #954, opened 2020-04-23
Clean up and fix faction weapon config

pledbrook:
I've made several changes here:

 1. We no longer create Bullpup and Vektor weapon templates to override the base game ones, but modify the existing ones.

 2. Fixed the inventory icons for some of the faction weapons.

 3. Moved some of the INI settings into the correct section.

This should hopefully fix the problem with faction weapons doing a lot of environmental damage, but not sure if that's the case.


LongWarOfTheChosen/Src/LW_WeaponsAndArmor/Classes/X2Item_FactionWeapons.uc
```diff
@@ -762,7 +169,7 @@ static function X2DataTemplate CreateBullpup_Coil_Template()
 	Template.EquipSound = "Magnetic_Weapon_Equip";
 	Template.Tier = 4;
 
-	Template.RangeAccuracy = default.SKIRMISHER_COIL_RANGE;
+	Template.RangeAccuracy = default.SKIRMISHER_SMG_RANGE;
```
Grobobobo, line 172: Im general, isn't it a standard to have each weapon range per weapon, and not per weapon type?

LongWarOfTheChosen/Config/XComGameData_WeaponData.ini
```diff
@@ -2178,10 +2119,10 @@ TEMPLAR_PISTOL_RANGE[23] = -80
 TEMPLAR_PISTOL_RANGE[24] = -90
 TEMPLAR_PISTOL_RANGE[25] = -100
 
-+Bullpup_Laser_ImagePath="UILibrary_LW_LaserPack.LaserSMG_Base"
-+Bullpup_Coil_ImagePath="UILibrary_LW_Overhaul.InventoryArt.CoilSMG_Base"
-+Vektor_Laser_ImagePath="CrossbowVektor.Inv_CrossbowVektor"
-+Vektor_Coil_ImagePath="CrossbowVektor.Inv_CrossbowVektor"
+Bullpup_Laser_ImagePath="UILibrary_LW_LaserPack.LaserSMG_Base"
+Bullpup_Coil_ImagePath="UILibrary_LW_Overhaul.InventoryArt.CoilSMG_Base"
```
Grobobobo, line 2123: Wait, THAT's what fixes the faction weapon icons?

LongWarOfTheChosen/Src/LW_WeaponsAndArmor/Classes/X2Item_FactionWeapons.uc
pledbrook: Seems like, although I don't know why. There's an awful lot of duplicate config. I don't even see a good reason to have different ranges for different weapon tiers. 🤷 For consistency, I'll split the Skirmisher SMG range into tiers and just do a lot of copying and pasting.

LongWarOfTheChosen/Config/XComGameData_WeaponData.ini
pledbrook: If you're talking about the removal the of `+` at the start of the line, no. That's just my obsessive removal of unnecessary `+`s. The fix for faction weapon icons is simply using the correct image path for inventory, for example on [line 729](https://github.com/long-war-2/lwotc/pull/954/files#diff-1a46b2cb210700088dfc5f63191b7e77R729). It looks like the laser and coil versions of the Bullpup are already configured correctly.

LongWarOfTheChosen/Src/LW_WeaponsAndArmor/Classes/X2Item_FactionWeapons.uc
pledbrook: Actually, I've changed my mind. I'd like to move to a single range config for all tiers of a weapon or range type. It will make updating range tables more reliable, since only one needs to be updated instead of five. There is always the option to add tier-specific range tables when we actually want a tier to have different ranges, but I'm just not sold on that being a good idea (except perhaps for plasma, which has its unique bonuses).

PR #971, opened 2020-05-03
Missed word "global"

FlashVanMaster:
Players could misunderstand which force level is told. Global or regional

### DIFF
--- a/LongWarOfTheChosen/Localization/BetterSecondWaveSupport.rus
+++ b/LongWarOfTheChosen/Localization/BetterSecondWaveSupport.rus
@@ -1,6 +1,6 @@
 ; LWOTC Translated
 [EnableChosen SecondWaveOptionObject]
-Tooltip="Активирует Избранных, когда сила Адвента становится 5."
+Tooltip="Активирует Избранных, когда глобальная сила Адвента становится 5."
 Description="Избранные: Активирует Избранных."
 ; End Translated
 


pledbrook:
Is this necessary? There is no difference between regional and global force level in LWOTC (nor in LW2).

FlashVanMaster:
In lw2 every region has also "force". (idk how it was callled in EN version)
Сохраненное изображение 2020-5-10_14-11-39 553
1 - is Advent Force
2 - i still idk what is RU variant is hidden there (but i know its vigilance)
3 - i guess its world can be translated as Power

So player can misunderstand power and advent force

FlashVanMaster:
Also i can change word. But i do not know exactly what word is hidden there (3), only guess. But it would be still very similar meaning. I think "global" will help to understand what is it exactly

pledbrook:
1: ADVENT strength (how many legions are in the region)
2: V for Vigilance
3: F for Force Level

There's no Russian translation for those letters as far as I can tell, so I assume the V and F are being converted directly to Cyrillic.

They can be changed in *LW_Overhaul.rus*, section `[UIStrategyMapItem_Region_LW]`, property name `m_strAlertLevel`.

aleksandr7768:
Then Russian translation:
Б for Vigilance
У for Force Level

m_strAlertLevel="Сила \"Адвента\": <XGParam:IntValue0> (Б:<XGParam:IntValue2> У:<XGParam:IntValue1>)"

pledbrook:
This has actually got me wondering whether we should refrain from using Force Level in our help text, since there are many players that simply won't know what it is. "Enemy tech level" is probably more accurate as a description.

PR #977, opened 2020-05-04
Replace Detailed Soldier Lists with LW2 version

pledbrook:
I've updated the old LW2 soldier list to work with WOTC. This gives us more control over what gets displayed in what order. Also means we don't have to rely on an integrated version of a Workshop mod.

In addition, I've borrowed some code from RPGO (thanks to Musashi!) that only overrides the soldier list item class if it hasn't already been overridden. This means that a player can enable Detailed Soldier Lists if they'd prefer to use that.


pledbrook:
It does, yes. There's not enough space to fit everything in 😆

pledbrook:
I think I still want to change how combat intelligence is displayed.

pledbrook:
Merged for dev build 20.

PR #981, opened 2020-05-06
Fix to Make Contact aspect of Issue # 815 + other strategy map UI items.

kdm2k6:
NOTE : This only modifies UIStrategyMapItem_Region_LW with a bit of code. It shows other files have changed; however, this must be because I branched off my other changes (since I wanted to test everything with a controller).

Fix to Make Contact aspect of Issue # 815.

It also fixes a number of situations in which strategy map UI items (related to regions) were unselectable. Finally, spaces have been replaced with tabs.

There are a few additional issues which I have solved, but will wait until this goes through :

1.] Some region UI items do not show their 3D mesh (viewable only with a controller / not m&k - so not testable with m&k). This can either be solved by modifying or hooking XComGameState_WorldRegion --> GetStaticMesh or through a workaround whereby I override several functions within UIStrategyMapItem_Region_LW and change 1 line of code, calling my own code rather than XComGameState_WorldRegion --> GetStaticMesh.

2.] Help icons, via flash, appear when regions are selected; unfortunately, these were never designed for Long War 2 usage, so they display the wrong icon and sometimes overlap. The best option is to 'null' them in flash via the movie controller - custom 'icon UIpanels' could always be added in later if need be; something which I have  more control over.


LongWarOfTheChosen/Config/XComLW_UI.ini
```diff
@@ -0,0 +1,63 @@
+[LW_Overhaul.UIResistanceManagement_LW]
+; Long War 2 defaults : 
+; 1.] Header button height = 26 
+; 2.] Header button font size = 20
+
+HEADER_BUTTON_HEIGHT_MK = 26				; Mouse & keyboard settings
```
pledbrook, line 6: These were in an earlier pull request, right?

pledbrook:
Ah, OK. This has the commits from the other pull request as well. Hmmm. I generally prefer to squash merge pull requests that have more commits than necessary, but that will cause issues for this pull request. Do you know about rebasing and how to use it?

kdm2k6:
If I'd have known it would grab my previous commits, I would have just created a branch off beta 3, made the changes to UIStrategyMapItem_Region_LW, then submitted. Can you just grab the changes from UIStrategyMapItem_Region_LW in this pull request? If not, I might just delete this one and make a new pull request off of beta 3 with the few code changes (there aren't many).

PR #984, opened 2020-05-07
Fix to Make Contact aspect of Issue # 815.

kdm2k6:
It also fixes a number of situations in which strategy map UI items (related to regions) were unselectable. Finally, spaces have been replaced with tabs.

There are a few additional issues which I have solved, but will wait until this goes through :

1.] Some region UI items do not show their 3D mesh (viewable only with a controller / not m&k - so not testable with m&k). This can either be solved by modifying or hooking XComGameState_WorldRegion --> GetStaticMesh or through a workaround whereby I override several functions within UIStrategyMapItem_Region_LW and change 1 line of code, calling my own code rather than XComGameState_WorldRegion --> GetStaticMesh.

2.] Help icons, via flash, appear when regions are selected; unfortunately, these were never designed for Long War 2 usage, so they display the wrong icon and sometimes overlap. The best option is to 'null' them in flash via the movie controller - custom 'icon UIpanels' could always be added in later if need be; something which I have more control over.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIStrategyMapItem_Region_LW.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIStrategyMapItem_Region_LW.uc
@@ -7,27 +7,6 @@
 
 class UIStrategyMapItem_Region_LW extends UIStrategyMapItem_Region;
 
-//var StaticMesh RegionMesh;
-//const NUM_TILES = 3;
-//var StaticMeshComponent RegionComponents[NUM_TILES];
-//
-//var public localized String m_strLockedTT;
-//var public localized String m_strUnlockedTT;
-//var public localized String m_strContactTT;
-//var public localized String m_strOutpostTT;
-//var public localized String m_strControlTT;
-//
-//var public localized String m_strScanForIntelLabel;
-//var public localized String m_strScanForOutpostLabel; 
-//var public localized String m_strButtonMakeContact;
-//
-//var UIButton ContactButton; 
-//var UIButton OutpostButton;
-//var UIPanel BGPanel;
-//var UIScanButton ScanButton;
-//
-//var array<float> CumulativeTriangleArea;
-
 var localized string m_strOutpostTitle;
 var localized string m_strAlertLevel;
 var localized string m_strLiberatedRegion;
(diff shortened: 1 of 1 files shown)


LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIStrategyMapItem_Region_LW.uc
```diff
@@ -7,37 +7,68 @@
...
 
+// KDM : Helper function which fills in variables that relate to a regions contact/radio relay status.
+// This has been made a function as it will likely be used in several places in the near future.
+simulated function UpdateRegionStatusVariables()
+{
+	CanMakeContact = class'UIUtilities_Strategy'.static.GetXComHQ().IsContactResearched() && (GetRegion().ResistanceLevel == eResLevel_Unlocked);
+	CanScanContact = GetRegion().bCanScanForContact;
+	CanMakeRadioRelay = class'UIUtilities_Strategy'.static.GetXComHQ().IsOutpostResearched() && (GetRegion().ResistanceLevel == eResLevel_Contact);
+	CanScanRadioRelay = GetRegion().bCanScanForOutpost;
+	RadioRelayInstalled = (GetRegion().ResistanceLevel == eResLevel_Outpost);
```
pledbrook, line 30: Why are these object variables rather than individual functions, like `CanMakeContact()`? I can't imagine that caching the information helps performance much, and now we have to worry about ensuring `UpdateRegionStatusVariables()` is called at the right times, otherwise the cached data could become out of sync with the real source of truth.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIStrategyMapItem_Region_LW.uc
```diff
@@ -7,37 +7,68 @@
...
+	if (CanMakeRadioRelay)
+	{
+		return true;
+	}
+
+	// KDM : If the region has a radio relay installed it is selectable.
+	if (RadioRelayInstalled)
+	{
+		return true;
+	}
```
pledbrook, line 65: I would consider replacing these multiple `if`s with a single line (requires implementation of the checks as functions:
```
return CanMakeContact() || CanMakeRadioRelay() || IsRadioRelayInstalled();
```

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIStrategyMapItem_Region_LW.uc
kdm2k6: Seems like a reasonable idea.

Will change them to functions.

kdm2k6:
Ok. I think I squashed, changed the commit description, and force pushed.

Hopefully .................

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIStrategyMapItem_Region_LW.uc
```diff
@@ -38,6 +17,48 @@ var localized string m_strMonthlyRegionalIncome;
...
+I have determined that :
+1.] Whether a region is the HQ or not should have no bearing on selectability so ignore it entirely.
+2.] The scan button type should have no bearing on selectability so ignore it entirely.*/
+simulated function bool IsSelectable()
+{
+	// KDM : The region is selectable if any of these conditions are true :
+	// 1.] It is contactable, regardless of whether contacting has commenced.
+	// 2.] It has been contacted, and a radio relay can be built, regardless of whether this building process has actually commenced.
+	// 3.] It has a radio relay installed.
+	if (CanMakeContact() || CanMakeRadioRelay() || IsRadioRelayInstalled())
```
pledbrook, line 52: You can just return the value of this condition (see my earlier comment). `if` branches that *just* return `true` or `false` and do nothing else can generally be compressed into a single `return`.

kdm2k6: Done.

kdm2k6:
Excellent. Thanks. Have a few questions about a couple other things, but will ask you tomorrow via Discord in all likelihood - I'm in Vancouver so it's getting late.

Talk to you tomorrow.

Issue #961, opened 2020-04-27
Controller support - can't select infiltration options


dadjumper:
Overall, controller works as it does in base WOTC, which I'm pleased with. 
One thing I've noticed you can't do is select the three options that come up once an infiltration is complete - "Boost infiltration," "View Squad" or "Abort."

They can be highligted with a controller, but as the A button is set to "Launch Mission," trying to select one of these options just launches the mission instead.

kdm2k6:
I've noticed that, it appears, someone is working on controller implementation (I see "bind controller R3 to PlaceDelayedEvacZone" in the commit history) other than myself.

Therefore, just letting everyone know I'm working on this particular issue and am nearing completion.

dadjumper:
Good to hear! I just noticed that the controller also can't select regions in the resistance management screen. Didn't want to open another issue so

pledbrook:
I think that's covered by [this change](https://github.com/long-war-2/lwotc/pull/963) if I understand you correctly.

pledbrook:
This was fixed by [commit 
53475b9](https://github.com/long-war-2/lwotc/commit/53475b95faf216fd52566bd754bf0f7c9c0a83d3)

destroyer11687:
I understand this issue has been closed but I downloaded the commit and placed the adjusted files in my lwotc folder, but I’m still having this issue. @pledbrook is there something I’m not doing right to get this to work?

PR #994, opened 2020-05-10
Fix for #961 - Controller can't select infiltration options.

kdm2k6:
This also allows mouse and keyboard users to use the arrow keys to select infiltration options and the spacebar / enter keys to "click".

-----------------------------------------

Modifies UIUtilities_LW --> BuildMissionInfoPanel().

When you click on a mission which is being infiltrated the infiltration screen, UIMission_LWLaunchDelayedMission, pops up. The function BuildMissionInfoPanel() creates the special mission information panel seen near the top of this screen. The problem is that within this function, a UIPanel and UIX2PanelHeader are created; however, they are never removed from the navigation system and end up causing problems with controller and arrow key navigation. Consequently, bIsNavigable has been set to false for both of them so they are no longer added to the navigation system.

-----------------------------------------

Modifies UIStrategyMapItem_Mission_LW --> OnMouseEvent() and OnUnrealCommand().

Originally, when a controller user clicked on the A button while a map item mission was highlighted, it would open a normal mission screen even if the mission was currently undergoing infiltration. As a result, OnUnrealCommand() had to be extended so as to call the same code as found in OnMouseEvent() --> FXS_L_MOUSE_UP. This code calls OpenInfiltrationMissionScreen(), which pushes an infiltration screen onto the screen stack if infiltration is occurring.

-----------------------------------------

Modifies UIMission_LWLaunchDelayedMission.

The main changes occur in BindLibraryItem(), RefreshNavigation(), and OnUnrealCommand().

1.] BindLibraryItem() has been overridden so that all of the UIButtons ("BOOST INFILTRATION", "VIEW SQUAD", "ABORT", "LAUNCH MISSION", and "WAIT") are created in the same manner whether the user is using a controller or mouse & keyboard; the main change is that they all have the style eUIButtonStyle_NONE. This guarantees that each button displays itself as a normal button rather than a hotlink; since there are up to 5 selectable buttons on this screen, controller hotlinks simply won't work.

2.] RefreshNavigation() has been overridden since the UIMission version is a bit of a mess, and to make it compatible with a controller and the arrow keys. It now clears the navigation system, adds any visible buttons to the navigation system, and then selects an initial button depending upon which buttons are visible.

3.] OnUnrealCommand() has been overridden so that the A button and Enter/Spacebar key click on whichever button happens to be selected.

There has also been a bit of cleanup and a fix for a problem in which the IgnoreButton could spawn several times on top of itself.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIMission_LWLaunchDelayedMission.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIMission_LWLaunchDelayedMission.uc
@@ -47,6 +47,74 @@ simulated function InitScreen(XComPlayerController InitController, UIMovie InitM
 	BuildScreen();
 }
 
+// Part of the fix for issue #961. This function makes sure that the "BOOST INFILTRATION", "VIEW SQUAD", "ABORT", "LAUNCH MISSION", and "WAIT"
+// buttons are set up properly for controller and keyboard support. The most important thing it does is make sure the buttons appear as normal
+// buttons rather than unselectable hotlinks when the controller is active. This way we can have a single column of buttons which can be selected
+// via the D-Pad or arrow keys and "clicked" with the A button or Enter/Spacebar keys.
+//
+// KDM : UIMission --> BindLibraryItem() code has been copied and the following modifications have been made :
+// 1.] All UIButtons have been given the style eUIButtonStyle_NONE so they appear as normal buttons when using a controller.
+// 2.] The delegate, OnSizeRealized, is no longer set for any of the UIButtons since it pointed to the empty function OnButtonSizeRealized().
+// 3.] Buttons are setup identically whether using a controller or mouse & keyboard; more specifically, positioning, width, and gamepad icons
+//		are no longer set when using a controller.
+// 4.] Any navigation setup has been removed since it should all be dealt with in RefreshNavigation().
+// 5.] The Ignore button is now created in this function along with all the other UIButtons. For more information please read the comments in BuildOptionsPanel(). 
+simulated function BindLibraryItem()
+{
+	local Name AlertLibID;
+	
+	AlertLibID = GetLibraryID();
+	
+	if (AlertLibID != '')
+	{
+		LibraryPanel = Spawn(class'UIPanel', self);
+		LibraryPanel.bAnimateOnInit = false;
+		LibraryPanel.InitPanel('LibraryPanel', AlertLibID);
+
+		ButtonGroup = Spawn(class'UIPanel', LibraryPanel);
+		ButtonGroup.InitPanel('ButtonGroup', '');
+
+		// KDM : Boost infiltration button
+		Button1 = Spawn(class'UIButton', ButtonGroup);
+		Button1.ResizeToText = false;
+		Button1.InitButton('Button0', "", , eUIButtonStyle_NONE);
+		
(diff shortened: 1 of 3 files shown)


LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIMission_LWLaunchDelayedMission.uc
```diff
@@ -47,6 +47,74 @@ simulated function InitScreen(XComPlayerController InitController, UIMovie InitM
...
+		Button2 = Spawn(class'UIButton', ButtonGroup);
+		Button2.ResizeToText = false;
+		Button2.InitButton('Button1', "", , eUIButtonStyle_NONE);
+
+		// KDM : Abort button
+		Button3 = Spawn(class'UIButton', ButtonGroup);
+		Button3.ResizeToText = false;
+		Button3.InitButton('Button2', "", , eUIButtonStyle_NONE);
+		
+		// KDM : The Confirm button is now given the style eUIButtonStyle_NONE so that, when using a controller, it doesn't appear as a hotlink.
```
pledbrook, line 92: Wouldn't this have been a problem in WOTC too? Or does the LWOTC launch screen cause specific issues?

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIMission_LWLaunchDelayedMission.uc
```diff
@@ -621,6 +699,122 @@ simulated function XComGameState_LWAlienActivity GetActivity()
...
+	{
+		return false;
+	}
+
+	bHandled = true;
+
+	switch(cmd)
+	{
+		// KDM : A button clicks on the selected button, if such a button exists.
+		// If no button was selected then fall down to the next case; this is to stay consistent with UIMission --> OnUnrealCommand().
```
pledbrook, line 782: Wait, doesn't this mean that Enter, Space, and A will close the screen if there is no button selected? I'm assuming that never happens, but it's weird behaviour for sure.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIMission_LWLaunchDelayedMission.uc
kdm2k6: Haven't played base XCom 2 in quite a while; however, based on recollection and looking at the code :

For something like a Guerrilla Ops, UIMission_GOps, the only numbered button to be used was Button1;  Button2 and Button3 are hidden in UpdateButtons(). Button 1 is a normal button which just displays the region or Guerrilla Ops title. Meanwhile, the confirm and ignore buttons are hotlinks which link to, something like, "Go to mission" and "Cancel mission".

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIMission_LWLaunchDelayedMission.uc
kdm2k6: Yep - That's why I added the comment to make sure I knew why this was done. In UIMission, if neither the confirm button nor Button1 are clicked, it just moves down and exits the screen if CanBackOut() is true.

The exact comment in UIMission is : "If you don't have a current button, fall down and hit the Navigation system". Not sure exactly when that would occur, but wanted to be consistent with it. If issues arise it is easy to add a break.

PR #998, opened 2020-05-10
Make UIOutpostManagement_ListItem compatible with RoboJumper's Rebel Job Selector.

kdm2k6:
Makes UIOutpostManagement_ListItem compatible with RoboJumper's Rebel Job Selector.

The following changes were made :

1.] JobHeaderX and JobHeaderWidth are now set again, within UIOutpostManagement_ListItem --> BuildItem(), since Robojumpers Rebel Job Selector makes use of them.

2.] Robojumpers Rebel Job Selector relies upon SetJobName() being called; however, in several instances this was no longer the case. Rather than setting the job name manually, I now feed the name back into SetJobName().

There has also been a bit of code cleanup - the job name for mouse & keyboard users is set in UpdateStaticData() since it does not change colour and only needs to be set once. The job name for controller users is still set in UpdateDynamicData() since it needs to be modified when a list item receives or loses focus.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIOutpostManagement_ListItem.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIOutpostManagement_ListItem.uc
@@ -71,6 +71,8 @@ simulated function BuildItem()
 	}
 
 	Width = List.Width;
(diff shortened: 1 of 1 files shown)


zyxpsilon:
Thank you!

PR #1005, opened 2020-05-14
Improve strategy map tooltips for Controller users.

kdm2k6:
1.] Provides a config variable, SHOW_REGION_TOOLTIPS_CTRL, which allows controller users to hide region tooltips; this is FALSE by default. The problem with region tooltips is that they can be quite long at times; furthermore, their information is largely irrelevant in terms of Long War.

Modifies : UIStrategyMapItem_Region_LW --> GenerateTooltip()

----------------------------

2.] Provides a config variable, SHOW_MISSION_TOOLTIPS_CTRL , which allows controller users to show mission tooltips; this is TRUE by default.
Long War provides useful mission tooltips to mouse & keyboard users when they mouse over mission icons on the bottom icon bar; unfortunately, this information is unavailable to controller users. With this new code, whenever a controller user hovers over a mission map item, a tooltip appears; this tooltip can include information like the mission's expiration time and its infiltration percentage.

Modifies : UIStrategyMapItem_Mission_LW --> UpdateFromGeoscapeEntity() and UIStrategyMapItem_Mission_LW --> GenerateTooltip()

GenerateTooltip() generates the tooltip while UpdateFromGeoscapeEntity() updates the tooltip information when appropriate.

----------------------------

3.] Whenever a controller user selects an infiltrating mission map item, its button becomes highlighted; unfortunately, Long War never modifies the text colour of its infiltration percentage string and, thus, it 'disappears'.

Modifies : UIStrategyMapItem_Mission_LW --> UpdateFromGeoscapeEntity()

If an infiltrating mission map item is selected, make sure the infiltration percentage string is black. If an infiltrating mission map item is not selected, make sure the infiltration percentage string is its normal blue.

----------------------------

Miscellaneous :

A new function, UIUtilities_Text_LW --> AddFontInfoWithColor(), was created which allows you to add a color, when manipulating a string.
An optional 4th parameter was added to X2EventListener_StrategyMap --> GetMissionSiteUIButtonToolTip() so Long War tooltip information could be easily accessed via a XComGameState_MissionSite.


LongWarOfTheChosen/Config/XComLW_UI.ini
```diff
@@ -60,3 +60,9 @@ LEVEL_ICON_OFFSET_CTRL = 0					; Controller settings
 LEVEL_ICON_SIZE_CTRL = 24
 LIST_ITEM_FONT_SIZE_CTRL = 24
 LIST_ITEM_FONT_SIZE_FANCY_CTRL = 26
+
+[LW_Overhaul.UIStrategyMapItem_Region_LW]
+HIDE_REGION_TOOLTIPS_CTRL = true			; Hides the region tooltips when using a controller. The problem with these tooltips is that they can be quite long and uninformative.
+
+[LW_Overhaul.UIStrategyMapItem_Mission_LW]
+SHOW_MISSION_TOOLTIPS_CTRL = true			; Shows tooltips for missions when using a controller. These tooltips are the same as what you would see on the bottom UI bar when using a mouse & keyboard.
```
pledbrook, line 68: I would prefer these to have a consistent name, so either `SHOW_...` or `HIDE_...`. I lean towards `SHOW_..._CTRL`.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIStrategyMapItem_Mission_LW.uc
```diff
@@ -337,6 +362,69 @@ simulated function OnLoseFocus()
 	super.OnLoseFocus();
 }
 
+// KDM : Tooltips need to be regenerated, on item focus, so that we get up to date information regarding mission expiration and infiltration percentage.
+//
+// It is IMPORTANT to note that the mission expiration and infiltration percentage will not update while the mission is selected; in other words, it will
+// not update when the user is inactive, but the geoscape time is slowly moving forward.
+//
+// The alternative was to regenerate the tooltip within UpdateFromGeoscapeEntity(), as this function is continuously called; however, I didn't want to
+// bombard the tooltip manager with too many update calls. Additionally, time creeps forward so slowly, when the user is inactive, that the mission
```
pledbrook, line 371: It creeps forward slowly, but if a player wanders away and then comes back a few minutes later, there will be a discrepancy in the infiltration percentages shown, correct? That may be a cost worth paying to avoid an even laggier Geoscape in the end game, but worth bearing in mind.

LongWarOfTheChosen/Config/XComLW_UI.ini
kdm2k6: Will do.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIStrategyMapItem_Mission_LW.uc
kdm2k6: Yes - it can create an inconsistency in display. I originally had the code in UpdateFromGeoscapeEntity(); however, I wasn't sure how expensive repeatedly recreating a tooltip was. Want to give it a try? If so, we can always change it back if any issues arise.

pledbrook: Yes, let's try having it in `UpdateFromGeoscapeEntity()` and hope we remember the potential cause if things seem laggy down the road 😆

kdm2k6: I've run into a bit of a snag - didn't work like I'd thought. Looking into it.

kdm2k6: Ok - I thought I had tooltips updating via UpdateFromGeoscapeEntity() beforehand, but I thought wrong. Finally figured out how to update it on the fly. Generally speaking, looks like tooltips are meant to be updated once, when an entity gains focus/selection.

Nonetheless, seems to work well now.

PR #1020, opened 2020-05-19
Fix finicky facility navigation for controllers + enable train officers staff slot for controllers

kdm2k6:
1.] Modifies : LW_OfficerPack_Integrated --> UIFacility_LWOfficerSlot --> OnClickStaffSlot().

When a UIStaffslot is clicked with the A button, OnUnrealCommand() simulates the click by calling OnClickStaffSlot(), and sending in FXS_L_MOUSE_UP. This case did not previously exist for the Long War slot; therefore, controllers users were unable to 'click it'. This has been fixed.

----------------------------------

2.] Added : LW_OfficerPack_Integrated --> UIScreenListener_Facility

Controller navigation, for UIFacilities, acts quite strangely at times. Here are some examples :
A.] If the rightmost staff slot is selected, and you hit D-Pad right, suddenly the room button, if one exists, becomes selected. It would be better if selection looped within the staff slot container.
B.] If you press the left or right bumper, the Avenger shortcuts, at the bottom of the screen, become selected. At this point, D-Pad down selects the room button, and D-Pad up does nothing.
C.] If you press right bumper --> D-Pad right --> D-Pad right, navigation becomes temporarily stuck on the 1st staff slot.
D.] If you press right bumper --> D-Pad right --> D-Pad left, there are suddenly 2 highlighted staff slots.

As a result, a screen listener has been implemented which, OnInit(), kills a facility's old navigation system, and creates a new, coherent setup. This applies to UIFacility and all subclasses of UIFacility, as they all share major components : a staff slot container, staff slots, a room container, and room buttons. This file has been commented extensively in order to give the viewer a good idea as to what, exactly, is being done and why.

Note that this file has been added to the LW_OfficerPack_Integrated package, since UIScreenListener_Facility_Academy_LWOfficerPack depends upon it, and according to UE3 documentation, if 2 files are in different packages, you can't dependOn a file in a package located after your own package.

----------------------------------

3.] Modifies : LW_OfficerPack_Integrated --> UIScreenListener_Facility_Academy_LWOfficerPack -- OnInit()

UIScreenListener_Facility_Academy_LWOfficerPack removes its staff slot container, creates a new one, then adds staff slots to it; this is to allow for officer slots to exist.

According to Robojumper, screen listener priority can not be set; this makes sense, as there are no variables of the sort within UIScreenListener. The problem is that my screen listener, UIScreenListener_Facility, relies upon the navigation system having been set up, so it can kill the old system, create a new one, and not have to worry about other screen listeners messing up the navigation system again.

Consequently, at the end of OnInit(), the navigation system is recreated and initial selection is dealt with; this guarantees that the navigation is set up properly no matter which screen listener runs 1st.

----------------------------------

4.] Spaces have also been converted to tabs in a few instances.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_OfficerPack_Integrated/Classes/UIFacility_LWOfficerSlot.uc
+++ b/LongWarOfTheChosen/Src/LW_OfficerPack_Integrated/Classes/UIFacility_LWOfficerSlot.uc
@@ -1,82 +1,76 @@
 //---------------------------------------------------------------------------------------
 //  FILE:    UIFacility_LWOfficerSlot.uc
-//  AUTHOR:  Amineri
-//           
+//  AUTHOR:  Amineri    
 //  PURPOSE: Reworked UIFacility_StaffSlot for officer functionality
 //---------------------------------------------------------------------------------------
 
-class UIFacility_LWOfficerSlot extends UIFacility_StaffSlot
-	dependson(UIPersonnel);
+class UIFacility_LWOfficerSlot extends UIFacility_StaffSlot dependson(UIPersonnel);
 
 var localized string m_strTrainOfficerDialogTitle;
 var localized string m_strTrainOfficerDialogText;
 var localized string m_strStopTrainOfficerDialogTitle;
 var localized string m_strStopTrainOfficerDialogText;
 
-//-----------------------------------------------------------------------------
 simulated function OnClickStaffSlot(UIPanel kControl, int cmd)
 {
+	local string StopTrainingText;
+	local XComGameState_HeadquartersProjectTrainLWOfficer TrainProject;
 	local XComGameState_StaffSlot StaffSlot;
 	local XComGameState_Unit UnitState;
-	//local XComGameState_HeadquartersXCom XComHQ;
-	local XComGameState_HeadquartersProjectTrainLWOfficer TrainProject;
-	local string StopTrainingText;
-
+	
 	StaffSlot = XComGameState_StaffSlot(`XCOMHISTORY.GetGameStateForObjectID(StaffSlotRef.ObjectID));
 
 	switch (cmd)
 	{
-	case class'UIUtilities_Input'.const.FXS_L_MOUSE_DOUBLE_UP:
-	case class'UIUtilities_Input'.const.FXS_L_MOUSE_UP_DELAYED:
-		if (StaffSlot.IsLocked())
-		{
-			ShowUpgradeFacility();
-		}
-		else if (StaffSlot.IsSlotEmpty())
-		{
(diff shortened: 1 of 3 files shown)


pledbrook:
When testing this, I couldn't interact with the rookie training slots, only the officer ones. I'll try to see why that's the case, but if you have any insight, that would be great. Also, sorry for leaving this for so long. I basically just took a sudden, long break from LWOTC.

kdm2k6:
Long time no see :]

It's been a long time so I can't remember much; however, my best guess is that it needs to be used in combination with one of my highlander changes (https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/issues/created_by/kdm2k6) which still haven't been closed after ... all ... these ... months ...

Perhaps it is this issue : https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/issues/866.

kdm2k6:
I was looking through my workshop and have 3 unpublished mods.

- One fixes the recruit screen (I never published it because the non-working LWotC one has to be removed before it will work).
- One modifies the "Manage Equipment Menu" for LWotC. It allows you to unequip various items via the squad select screen, using a controller.
- One fixes LeaderEnemyBoss's "No Drop Down Lists" mod so it is usable with controllers. I wonder if this is what was needed. I won't publish it yet - but here is the mod as an attachement.
[2098062078.zip](https://github.com/long-war-2/lwotc/files/5411890/2098062078.zip)

pledbrook:
Yeah, the rookie Recruit screen is a bit of a nightmare. It's not scrollable if the number of recruits is too large for the list container.

pledbrook:
OK, it seems your fix to LEB's No Drop Down Lists is needed.

pledbrook:
I've asked on the mod's [Steam page](https://steamcommunity.com/sharedfiles/filedetails/?id=865667559) whether LEB is willing to let us either integrate the mod or publish a different version with a fix for controllers. I don't think LEB is active in XCOM 2 any more, but they are active in another modding community, so may respond. If there is no response, I think we can look at integrating the mod.

kdm2k6:
I emailed him months back, just before I made the mod. However, I received no response.

pledbrook:
Ah, OK. Do you want to release a separate mod or integrate it into LWOTC? I don't know how useful the mod is outside the context of Long War. Actually, the main point is it's usefulness for any Not Created Equal implementation, which isn't exclusive to Long War. So a separate (fixed) mod makes sense.

kdm2k6:
I had originally intended to publish it as a separate mod, so that's definitely ok with me. I already have it uploaded, just need to hit the "publish" button.

I'll wait for a bit and see if he responds to you first. Although, I doubt he will have an issue with it.

PR #1023, opened 2020-05-20
Make dark event list items controller compatible

kdm2k6:
Modifies : UIAdventOperations_LW --> GetDarkEventListUI()

The dark event list has been moved up, so as to align with the top of the dark event card to its left. Additionally, the list was made navigable so that controller users could navigate the dark event list items.

-------------------------

Modifies : UIDarkEventListItem --> OnReceiveFocus()

When a controller user navigates to a different dark event list item, we need the dark event card to update; previously this was only dealt with in mouse & keyboard code.

-------------------------

A bunch of spacing issues were dealt with as well.


pledbrook:
Thanks for the contribution!

PR #1034, opened 2020-05-24
Make Officer Promotion Screen controller compatible + a few related fixes.

kdm2k6:
Modifies : UIArmory_LWOfficerPromotion & UIArmory_LWOfficerPromotionItem

1.] The officer training screen has now been made controller capable; D-Pad up/down changes rows, while D-Pad left/right changes abilities. Additionally, ability selection remains consistent when changing rows; if the left ability is selected in the old row, it will also be selected in the new row; same goes for the right ability.
2.] The 'leadership history' button has been made controller capable, via a hotlink.
3.] The navigation system was also modified, for controller users, such that it would only update when necessary. The reason this was done was that each time the navigation system updates, it flickers, since it needs to clear itself, re-create itself, then show itself. 
4.] Spaces --> Tabs

------------------------------------

Modifies : UIAlert_LWOfficerTrainingComplete

When officer training finishes, generally speaking, 2 buttons should appear; 1 button allows you to view the soldier, while the other button simply 'carrys on'. Several fixes were needed :
1.] Both the controller A and B buttons called OnCancelClicked(), equivalent to "carry on"; this was fixed in OnUnrealCommand().
2.] For mouse & keyboard users, these two buttons were placed on top of each other due to a call to OnTrainingButtonRealized(); subsequently, this call was removed.

------------------------------------

Modifies : UIScreenListener_LWOfficerPack --> OnOfficerButtonCallback()

Previously, the officer pawn could be rotated when accessing the officer screen through the officer slot, but not when accessing it through the 'Officer Abilities' menu button. Turns out it was missing a call to CreateSoldierPawn().

### DIFF
--- a/LongWarOfTheChosen/Src/LW_OfficerPack_Integrated/Classes/UIAlert_LWOfficerTrainingComplete.uc
+++ b/LongWarOfTheChosen/Src/LW_OfficerPack_Integrated/Classes/UIAlert_LWOfficerTrainingComplete.uc
@@ -1,29 +1,28 @@
 //---------------------------------------------------------------------------------------
-//  FILE:    UIAlert_LWOfficerTrainingComplete.uc
+//  FILE:    UIAlert_LWOfficerTrainingComplete.uc 
 //  AUTHOR:  Amineri
 //  PURPOSE: Customized UI Alert for Officer training 
-//           
 //---------------------------------------------------------------------------------------
 class UIAlert_LWOfficerTrainingComplete extends UIAlert;
 
-//override for UIAlert child to trigger specific Alert built in this class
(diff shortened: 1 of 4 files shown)


pledbrook:
Thanks, this looks good. I can't merge it until I've managed to merge the fix for GTS staff slots, though.

By the way, when officer training completes and the player selects the "go to soldier" option, the right stick doesn't work in the officer training screen to rotate the pawn. Note that this *does* work when accessing the officer training screen from the GTS officer training slot. However, I tested it with mouse and keyboard and I couldn't rotate the pawn then, either. So probably something independent of whether a controller is being used or not.

PR #1041, opened 2020-05-26
Improve New Promotion Screen controller compatibility + a few related fixes

kdm2k6:
Modifies : NPSBDP_UIArmory_PromotionHero --> ChangeSelectedColumn

Beforehand, controller navigation of the promotion screen was not ideal; whenever column selection was changed, the 1st row of the new column was selected regardless of the previous column's row selection. Now there is an attempt to maintain the current 'row selection' when selecting different columns; if this is not possible, due to different column sizes, the closest possible row is selected.

----------------------------

Modifies : NPSBDP_UIArmory_PromotionHero --> ConfirmAbilityCallbackWithTracking

Originally, an ability's focus was lost upon being selected, due to the promotion data being re-populated. This has now been fixed.

----------------------------

Modifies : NPSBDP_UIArmory_PromotionHero --> InitPromotion

For some reason the scrollbar was hanging off the main promotion panel, and was oftentimes hidden by the soldier pawn. This has been rectified.

----------------------------

Modifies : NPSBDP_UIArmory_PromotionHero --> UpdateNavHelp

The 'show abilities' navigation tip has been moved to the left side, from the center, so it doesn't collide with other navigation tips.

### DIFF
--- a/LongWarOfTheChosen/Src/NewPromotionScreenByDefault_Integrated/Classes/NPSBDP_UIArmory_PromotionHero.uc
+++ b/LongWarOfTheChosen/Src/NewPromotionScreenByDefault_Integrated/Classes/NPSBDP_UIArmory_PromotionHero.uc
@@ -30,6 +30,43 @@ var int Position, MaxPosition;
 
 var int AdjustXOffset;
 
+simulated function ChangeSelectedColumn(int oldIndex, int newIndex)
+{
+	local int i, NewColumnAbilityIndex, NewColumnAbilities, OldColumnAbilityIndex;
+	local UIArmory_PromotionHeroColumn OldColumn, NewColumn;
+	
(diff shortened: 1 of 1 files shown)


pledbrook:
This seems to be working well, thanks. However, the scroll bar is currently running right down the middle of the MSGT abilities, which is less than ideal. Is this what you intended?

kdm2k6:
That's interesting as it seems to work on my computer using only the highlander and the version of LWotC from a while back (when I was working on it).

I did find one of my comments in my change which states : 
// KDM : The X position was originally set to -465; however, I am perplexed as to why, since the scrollbar was much too far to the right. In fact it was nearly always blocked by the soldier pawn.

Perhaps something changed between when I worked on it and now, although I'm not entirely sure.

PR #1042, opened 2020-05-26
Detailed Soldier List Item D-Pad Down fix

kdm2k6:
Modifies : UIPersonnel_SoldierListItemDetailed

When a soldier list was loaded, the 1st D-Pad down press did nothing; strangely enough, this was not a problem with D-Pad up. It turns out that UIIcons, whose navigation is ON by default, weren't being removed from the navigation system. Consequently, navigation was removed from : APIcon, BadTraitIcon, and BonusAbilityIcon.


pledbrook (approved):
Looks good to me, thanks!

PR #1046, opened 2020-05-27
Fix Detailed Soldier Will & PCS text colour when focused.

kdm2k6:
Modifies : UIPersonnel_SoldierListItemDetailed --> AddClassColumnIcons

Previously, the PCS stat value would not display black text when the list item was selected; consequently it was very difficult to read. The reason for this was that a call to GetStatBoostString(), which called GetColoredText(), was wrapped in a call to class'UIUtilities_Text'.static.GetColoredText(). Wrapping text in 2 font tags creates unpredictable outcomes; therefore, GetStatBoostString() was modified, to take list item selection into account, and the outer call to class'UIUtilities_Text'.static.GetColoredText() was removed.

Modifies : UIPersonnel_SoldierListItemDetailed --> GetStatBoostString

GetStatBoostString now takes a 'focused' parameter so that it can determine which colour to display the PCS stat value, if one exists.
1.] Disabled list items always display a greyish text.
2.] Focused list items, if not disabled, always display black text.
3.] If 1 & 2 are false, green is used for values greater than 0 and red is used for values less than 0.

Modifies : UIPersonnel_SoldierListItemDetailed --> UpdateItemsForFocus

Will values will now display in black if the list item is focused, making things much easier to read.

### DIFF
--- a/LongWarOfTheChosen/Src/DetailedSoldierListWOTC_Integrated/Classes/UIPersonnel_SoldierListItemDetailed.uc
+++ b/LongWarOfTheChosen/Src/DetailedSoldierListWOTC_Integrated/Classes/UIPersonnel_SoldierListItemDetailed.uc
@@ -756,9 +756,9 @@ function ShowDetailed(bool IsDetailed)
 	}
 }
 
-simulated function string GetStatBoostString(XComGameState_Item ImplantToAdd)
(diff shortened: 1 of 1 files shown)


pledbrook (approved):
Thanks, looks good. I prefer "Focused" over "Focussed", but I notice that the class already uses the latter, so it's fine.

PR #1053, opened 2020-06-01
Add ability to Strip Weapon Upgrades for Controllers + few weapon upgrade fixes

kdm2k6:
Modifies : XComGameState_LWListenerManager

The 3 functions : AddArmoryStripWeaponsButton(), OnStripWeaponClicked(), ConfirmStripSingleWeaponUpgradesCallback() were removed because :
1.] They were never used.
2.] They led to confusion when investigating what changes needed to be made, in order to add 'strip weapon upgrade' functionality for controllers.

----------------------------------------

Modifies : X2EventListener_Headquarters

1.] Added a new function, CreateXComArmoryListeners(), which adds event listeners dealing with the Armory on the Avenger.

2.] Added an event listener function, OnWeaponUpgradeNavHelpUpdated(), which is triggered by the event 'UIArmory_WeaponUpgrade_NavHelpUpdated'; this is called whenever the navigation help system updates within UIArmory_WeaponUpgrade.

OnWeaponUpgradeNavHelpUpdated() adds a 'Strip All Weapon Upgrades' tip to the navigation help system for both mouse & keyboard users as well as controller users. Additionally, it adds a 'Strip Weapon Upgrades' tip to the navigation help system for controller users; this tip is not needed for mouse & keyboard users since they have their own list item.

----------------------------------------

Modifies : UIScreenListener_ArmoryWeaponUpgrade_LW

1.] Removed a variety of code which was either unused or no longer needed; this includes the UIButton StripWeaponUpgradesButton, the boolean AddedNavHelp, as well as the functions: RegisterForEvents(), OnWeaponUpgradeListChanged(), and UpdateNavHelp(). More precise reasoning is described in the files header.

2.] Added the function OnWeaponUpgradeCommand(), which intercepts UIArmory_WeaponUpgrade's input via the highlander function SubscribeToOnInputForScreen(). This was done for several reasons :

- It allows left stick click to be mapped to 'Strip all weapon upgrades'.
- It allows right stick click to be mapped to 'Strip [this] weapon upgrades'.
- It allows for SlotsList selection to be saved and restored when a controller is being used. This is important because it allows controllers users to, say, click on the 3rd weapon slot, select a weapon upgrade, and then have the 3rd weapon slot be selected upon return. Previously, the 1st weapon slot would "always" be selected upon returning from choosing a weapon upgrade.

3.] RefreshScreen() was modified so that the text for 'Strip Weapon Upgrades' did not disappear when the list item gained focus.

4.] A few other minor issues and cleanup, including spaces to tabs, were made.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2EventListener_Headquarters.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2EventListener_Headquarters.uc
@@ -18,6 +18,7 @@ static function array<X2DataTemplate> CreateTemplates()
 	local array<X2DataTemplate> Templates;
 
 	Templates.AddItem(CreateXComHQListeners());
+	Templates.AddItem(CreateXComArmoryListeners());
 	Templates.AddItem(CreateCovertActionListeners());
 	Templates.AddItem(CreateWillProjectListeners());
 
@@ -37,6 +38,19 @@ static function CHEventListenerTemplate CreateXComHQListeners()
 	Template.AddCHEvent('CanTechBeInspired', CanTechBeInspired, ELD_Immediate, GetListenerPriority());
 	Template.AddCHEvent('UIAvengerShortcuts_ShowCQResistanceOrders', ShowOrHideResistanceOrdersButton, ELD_Immediate, GetListenerPriority());
 
+	Template.RegisterInStrategy = true;
+
+	return Template;
+}
+
+// KDM : Event listeners dealing with the Armory on the Avenger.
+static function CHEventListenerTemplate CreateXComArmoryListeners()
+{
+	local CHEventListenerTemplate Template;
+
+	`CREATE_X2TEMPLATE(class'CHEventListenerTemplate', Template, 'XComArmoryListeners');
+	Template.AddCHEvent('UIArmory_WeaponUpgrade_NavHelpUpdated', OnWeaponUpgradeNavHelpUpdated, ELD_Immediate, GetListenerPriority());
+	
 	Template.RegisterInStrategy = true;
 
 	return Template;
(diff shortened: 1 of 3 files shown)


LongWarOfTheChosen/Src/LW_Overhaul/Classes/XComGameState_LWListenerManager.uc
```diff
@@ -401,66 +407,6 @@ function EventListenerReturn AddSquadSelectStripWeaponsButton (Object EventData,
 	return ELR_NoInterrupt;
 }
 
-function EventListenerReturn AddArmoryStripWeaponsButton (Object EventData, Object EventSource, XComGameState GameState, Name InEventID, Object CallbackData)
```
pledbrook, line 404: Please don't remove these functions if they are still referenced by commented-out code in `InitListeners()`. I never finished the migration away from `XCGS_LWListenerManager`. What's left may still need to be migrated/implemented.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2EventListener_Headquarters.uc
```diff
@@ -77,6 +91,64 @@ static protected function int GetListenerPriority()
...
+			ListenerManager.OnStripUpgrades, false, class'UIUtilities_LW'.default.m_strTooltipStripWeapons);
+	
+		// KDM : 'Strip Weapon Upgrades' is a CustomizeList list item for mouse & keyboard users, so it doesn't need to be added to the
+		// navigation help system.
+		if (`ISCONTROLLERACTIVE)
+		{
+			// KDM : Right stick click corresponds to 'Strip Weapon Upgrades'
+			NavHelp.AddLeftHelp(CAPS(class'UIScreenListener_ArmoryWeaponUpgrade_LW'.default.strStripWeaponUpgradesButton), 
+				class'UIUtilities_Input'.const.ICON_RSCLICK_R3, , false, 
+				class'UIScreenListener_ArmoryWeaponUpgrade_LW'.default.strStripWeaponUpgradesTooltip);
```
pledbrook, line 145: The text for these two is so long, that it runs beneath the LB/RB hint for cycling through soldiers. Should one of them be put at the end? Or shorter text developed? Perhaps just "Strip Upgrades" and "Strip upgrades from all weapons"? In fact, the second is more descriptive than the existing text, but it's even longer. Perhaps it should go on the right of the screen?

pledbrook (changes requested):
Good work, thanks. I think we do need a better solution for the "Strip Weapon..." hints and as requested, please restore the LWListenerManager code.

PR #1099, opened 2020-08-11
Add config option for whether steadying weapon breaks concealment.

Davi-DeGanne:
Default behavior should remain unchanged, since the default is that steadying *does* break concealment. This just allows users to revert back to the original LW2 behavior if they so choose, without needing to rebuild from source.

(I have never done X2 modding before, I simply examined the way similar things were already done, and copied syntax/procedures from there. This may be a small PR but please take extra care when checking my work.)


Davi-DeGanne:
I was informed in the discord that the stock/steady changes are being reverted to LW2 behavior so I guess this is unnecessary.

PR #1141, opened 2020-10-29
Reaper rework

pledbrook:
See commit for details


LongWarOfTheChosen/Config/XComGame.ini
```diff
@@ -102,10 +102,13 @@ DisplayName="Long War of the Chosen"
 +RequiredMods="robojumperSquadSelect_WotC"
 +RequiredMods="BetterSecondWaveSupport"
 +RequiredMods="WOTCRevertOverwatchRules"
++RequiredMods="WotCCombatKnives"
```
Grobobobo, line 105: I thought the only reason Why you'd make it a dependency is to not have to commit to putting experimental content in, but now? why not just integrate the assets you want in? Gives you a lot more control over things overall, and it's not that much effort either.

LongWarOfTheChosen/Config/XComGameData_WeaponData.ini
```diff
@@ -959,6 +959,49 @@ LMG_ALL_RANGE[24] = -20
...
+MID_LONG_ALL_RANGE[13] = 0
+MID_LONG_ALL_RANGE[14] = 0
+MID_LONG_ALL_RANGE[15] = 0
+MID_LONG_ALL_RANGE[16] = 0
+MID_LONG_ALL_RANGE[17] = 0
+MID_LONG_ALL_RANGE[18] = 0
+MID_LONG_ALL_RANGE[19] = 0
+MID_LONG_ALL_RANGE[20] = 0
+MID_LONG_ALL_RANGE[21] = 0
+MID_LONG_ALL_RANGE[22] = -10
```
Grobobobo, line 984: Now that i look at it, -10 per tile does sound extreme, I'd probably drop it to -5 or -3 aim per tile.

LongWarOfTheChosen/Config/XComClassData.ini
```diff
@@ -1737,15 +1740,14 @@ bHasClassMovie=true
 !SoldierRanks=()
 ; squaddie
 +SoldierRanks=(AbilitySlots=((AbilityType=(AbilityName="Shadow", ApplyToWeaponSlot=eInvSlot_Unknown)), \\
```
Grobobobo, line 1742: Wait, so you've reworked their vektors to be like strike rifles, but didn't give them squadsight? that seems really strange.

Grobobobo (changes requested):
There doesn't seem to be anything wrong with the perk tree balance itself, the numbers will probably iron itself out with more testing. that said, i do have some issues with it, mentioned in comments.

LongWarOfTheChosen/Config/XComClassData.ini
pledbrook: Well, you only said the changes would help the weapon work *with* squadsight, and they get squadsight as a potential XCOM row perk to make use of those changes. I'm not ready to give all Reapers squadsight out of the gate.

I do wonder whether Squadsight should be on the weapons themselves, but that would be a larger change that wouldn't be appropriate for 1.0.

LongWarOfTheChosen/Config/XComGameData_WeaponData.ini
pledbrook: OK, -4 is it 😆

LongWarOfTheChosen/Config/XComGame.ini
pledbrook: I didn't want to go through the (unknown) effort of copying all the assets we need. I didn't think I could just copy the UPK, because it would conflict with any changes to the mod if someone wanted to use it with LWOTC. Having a dependency also makes it easy for us to effectively give credit to the mod author. And we benefit from improvements to the mods/animations/etc.

If it's easier than I think to integrate the assets, let me know.

LongWarOfTheChosen/Config/XComClassData.ini
Grobobobo: Then how about putting squadsight somewhere on the main tree?

LongWarOfTheChosen/Config/XComGame.ini
Grobobobo: Well, unless the mod has something I don't know about it should be as easy as putting the packages into our content and changing the references. In addition, straight up integrating is better for stability, remember how the dedicated pistol mod update broke things? maybe sometimes we don't need updates.

pledbrook:
Merged.

PR #1142, opened 2020-10-30
AI tweaks

NilchEi:
Is 2 commits due to fixing a typo that somehow sneaked in


Grobobobo (approved):
I do have some minor concerns, but it looks good for now.

PR #1145, opened 2020-11-07
Make the first SPARK easier to get

pledbrook:
This removes ADVENT Robotics as a requirement for Mechanized Warfare, so you don't have to acquire 3 drone wrecks, and the cost of Mechanized Warfare it much reduced. However, it does now require a MEC wreck to start the project. This adds an effective FL gate to getting your first SPARK.


Grobobobo (approved):
After long and diligent checking of this massive pull request, I conclude that there is no mistake anywhere, despite it's amazing size.

PR #1148, opened 2020-11-08
Naja AI Update

NilchEi:
Series of changes that significantly improve Naja's response to being flanked. Instead of just moving around the flanking soldier, now it tries to move as far away as possible without losing sight of XCOM, trying to position itself 22~27 tiles away throughout the battle if not provided with the chance to shoot. . Also built custom scamper profile with same movement logic which gave it similar distancing preference.

* Original AnyEnemyWithin12Tiles actually did not trigger even when there was nearby enemy, so changed it to more reliable scoring system.
* SniperMove profile modification: seeks minimum 22 tile distance with enemy visibility, low fCoverWeight so AI deviate less from ideal distance.
* SniperMove behaviour modification: uses RestrictedDestination to guarantee cover, allows dash movement for better sniper spot search and disengagement.

* ActionSelector Behaviours reworked to call SniperMove for every movement except HuntEnemyWithCover.
* Removed RandMoveIfEnemyClose so Naja does not randomly lose opportunity to make normal sniper shots; the behaviour did not work in original anyways as MoveIfAnyEnemyWithin12Tiles had not been functional. This also made NeedsReload redundant as TrySniperShootOrReload checks it.
* Child[6] and Child[7] of LWNajaRedFirstActionSelector allows Naja to track down XCOM that has moved out of sight.
* HoldGroundIfEnemyVisible prevents other backup movement profiles from SniperMove unnecessarily causing Naja to move away from ideal sniper spot.

* ScamperRoot_Naja added ~~with additional stuff to support it in ini only, as I'm unfamiliar with & cannot install XCOM2 devkit to make uc implementation.~~ to Naja charTemplate as requested.


LongWarOfTheChosen/Config/XComAI.ini
```diff
@@ -78,8 +78,17 @@
 +NewBehaviors=(BehaviorName=GenericAIRoot, NodeType=Selector, Child[0]=TryNonAggressiveBehavior, Child[1]=TryMindControlledRoot, Child[2]="::CharacterRoot", Child[3]=RandOverwatch50, Child[4]=DoCower, Child[5]=SkipMove)
 
 ; "GenericScamperRoot" - If the unit is unable to find a valid scamper tile, try again with hazards turned off.
+; Naja scamper enabled via ini, check added on Child 1 of GenericScamperRoot. Everything new behavior except ScamperRoot_Naja can be nuked after uc implementation.
 +BehaviorRemovals="GenericScamperRoot"
-+NewBehaviors=(BehaviorName=GenericScamperRoot, NodeType=Selector,       Child[0]=CounterBeaglerushManeuver,       Child[1]=MoveStandardWithLoSUnsafe,       Child[2]=MoveStandardUnsafe, Child[3]=MoveStandardWithLoSUnsafeIgnoreHazards, Child[4]=MoveStandardUnsafeIgnoreHazards, Child[5]=SkipMove)
++NewBehaviors=(BehaviorName=GenericScamperRoot, NodeType=Selector, Child[0]=CounterBeaglerushManeuver, Child[1]=NajaScamperCheck, Child[2]=MoveStandardWithLoSUnsafe, Child[3]=MoveStandardUnsafe, Child[4]=MoveStandardWithLoSUnsafeIgnoreHazards, Child[5]=MoveStandardUnsafeIgnoreHazards, Child[6]=SkipMove)
```
Grobobobo, line 83: That is absolutely not the way to do scamper. Instead, change the unit template's `CharTemplate.strScamperBT`

Grobobobo (changes requested):
Instead of doing the thing you did, add the line CharTemplate.strScamperBT = "ScamperRoot_Naja" in the character's template.

LongWarOfTheChosen/Config/XComAI.ini
NilchEi: Yeah, that's personal code I've been using since I can't compile .uc, probably was not a good idea to add it to pull request. I've removed the change and only left ScamperRoot_Naja behaviour.

Grobobobo (changes requested):
It's still merging to the wrong branch, go for master.

PR #1155, opened 2020-11-10
Naja movement AI update

NilchEi:
* Original AnyEnemyWithin12Tiles actually did not trigger even when there was nearby enemy, so changed it to more reliable scoring system.
* SniperMove profile modification: seeks minimum 14 tile distance with enemy visibility, low fCoverWeight so AI is willing to move to half-cover if it's only cover available far away.
* SniperMove behaviour modification: uses RestrictedDestination to guarantee cover, allows dash movement for better sniper spot search and disengagement.

* ActionSelector Behaviours reworked to call SniperMove for every movement except HuntEnemyWithCover.
* Removed RandMoveIfEnemyClose so Naja does not randomly lose opportunity to make normal sniper shots; the behaviour did not work in original anyways as MoveIfAnyEnemyWithin12Tiles had not been functional. This also made NeedsReload redundant as TrySniperShootOrReload checks it.
* Child[6] and Child[7] of LWNajaRedFirstActionSelector allows Naja to track down XCOM that has moved out of sight.
* HoldGroundIfEnemyVisible prevents other backup movement profiles from SniperMove unnecessarily causing Naja to move away from ideal sniper spot.

* ScamperRoot_Naja gives Naja scamper high preference towards 14~18 tiles distance

### DIFF
--- a/LongWarOfTheChosen/Config/XComAI.ini
+++ b/LongWarOfTheChosen/Config/XComAI.ini
@@ -378,8 +378,8 @@ MAX_SURPRISED_SCAMPER_PATH_LENGTH=7		; Scamper paths longer than this number of 
 
 ; ***                                           MOVEMENT PROFILES                                                   ***
 
-; snipermove that prioritize 22-27 distance
-+m_arrMoveWeightProfile=(Profile=SniperMove, fCoverWeight=0.1f, fDistanceWeight=2.0f, fFlankingWeight=0.0f, fEnemyVisWeight=1.0f, fEnemyVisWeightPeak1=1.0, fAllyVisWeight=0.1f, fCloseModifier=0.2f, fFarModifier=1.0f)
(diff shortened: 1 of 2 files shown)


Grobobobo (approved):
The most impactful change of it is that she will almost always shoot when NOT flanked, and that on scamper she will move away from an enemy rather then to the enemy, I don't think it's particularly balance breaking, since najas being disabled on being flanked is kept, and that's what most people use.

NilchEi:
Another important point to make is that I played a lot to ensure they still have high preference towards position where they have direct visibility on XCOM (so never beyond 19 tiles; see [here](https://media.discordapp.net/attachments/592382115228942349/776836226073362503/NajaScoring.png) for specifics). This means player is still able to target Naja with their attack/disables, and can flank them with dash movement except in rare cases where multiple floors are involved.

PR #1158, opened 2020-11-17
Fix Region/Haven controller selectability and associated 3D mesh.

kdm2k6:
Added X2EventListener_WorldRegion :

This class listens for WorldRegionGetStaticMesh and WorldRegionGetMeshScale, 2 events which are triggered within XComGameState_WorldRegion.

WorldRegionGetStaticMesh is used to make sure that, when using a controller and on the strategy map, each selectable region/haven is associated with a 3D mesh. Previously, Long War 2 introduced states which the base game did not handle; consequently, you could end up with selectable regions/havens with no 3D mesh.

WorldRegionGetMeshScale simply makes sure that the 3D mesh associated with a selectable region/haven is the correct size on the strategy map.

--------------------------------------

Modified UIStrategyMapItem_Region_LW :

UIStrategyMapItem_Region_LW.IsSelectable was previously modified to make sure that the following was true when using a controller :
1.] Contactable regions were selectable.
2.] Contactable regions, which were scanning for contact, were selectable.
3.] Havens, where a radio relay could be built, were selectable.
4.] Havens, with an installed radio relay, were selectable.

The one condition I missed was : Havens, where a radio relay could not be built due to tech level, should still be selectable.

Consequently, I have modified IsSelectable and simplified the logic so any "contactable" or "contacted" region/haven is now selectable.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIStrategyMapItem_Region_LW.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIStrategyMapItem_Region_LW.uc
@@ -24,14 +24,9 @@ simulated function bool CanMakeContact()
 	return (class'UIUtilities_Strategy'.static.GetXComHQ().IsContactResearched() && (GetRegion().ResistanceLevel == eResLevel_Unlocked));
 }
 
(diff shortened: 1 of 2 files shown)


pledbrook (approved):
Thanks! Looks good.

PR #1170, opened 2020-11-24
Soldier List Item D-Pad Down fix

kdm2k6:
This fix was originally implemented for a detailed soldier list item (see pull request #1042); however, the detailed soldier list has since been replaced with a modified LW2 list. Hence, this fix was needed.

----------------------

Modifies : UIScreenListener_LWOfficerPack.AddOfficerIcon_PersonnelListItem

1.] UIPersonnel_SoldierListItem_LW.AddAdditionalItems triggers the event 'OnSoldierListItemUpdate_End' which ends up calling AddOfficerIcon_PersonnelListItem; its goal is to add an officer icon to a soldier list item, if need be. Unforunately, while icons are added to the navigation system by default, they can create controller navigation issues. In this case, the first D-Pad down press, when viewing a soldier list, would appear to be ignored. As a result, the officer icon needed to be removed from the navigation system.

2.] The officer icon's bDisableSelectionBrackets was set to true, so that brackets no longer appear around the icon when the soldier is selected; these brackets were unnecessary and looked odd.

3.] A bunch of code cleanup, mainly tabs and spacing, was done within this file.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_OfficerPack_Integrated/Classes/UIScreenListener_LWOfficerPack.uc
+++ b/LongWarOfTheChosen/Src/LW_OfficerPack_Integrated/Classes/UIScreenListener_LWOfficerPack.uc
@@ -11,8 +11,10 @@ var localized string strCommandingOfficerTooltip;
 var localized string strSubordinateOfficerTooltipTitle;
 var localized string strSubordinateOfficerTooltip;
 
-var bool bRegisteredForEvents; // has this registered this game session?
-var bool bLastUpdateStrategy; // has this registered since a tactical/strategy switch ?
+// Has this registered this game session ?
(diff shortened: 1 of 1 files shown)


pledbrook (changes requested):
I'm sorry, but there are too many irrelevant formatting changes here. I don't mind (and in fact approve of) a few changes to tidy up code around where the relevant changes, but the bulk of a pull request should consist of such relevant changes.

I don't have the energy or concentration to uncover what's important/significant and what's not, so would you be able to redo this pull request please so it's more focused? Would be much appreciated.

pledbrook (approved):
Actually, I found the relevant changes and had tested it, so I will merge. It's a simple thing, but I really, really want this change :smile: However, in future, please try not to overwhelm pull requests with formatting changes.

PR #1176, opened 2020-11-26
Disable RecruitScreen Screen Listener so Recruit Screen is moddable

kdm2k6:
Comments out : UIScreenListener_RecruitSoldiers

The current Recruit Screen implementation is problematic for 2 reasons :

1.] When you recruit a soldier from the recruit screen, the screen reverts to its base game 'look'.
2.] Recruit attribute text does not change colour when different recruit rows are selected. This is because, while the screen listener implements OnReceiveFocus and OnLoseFocus, these are only called when the screen itself receives and loses focus; they are not called when individual list items receive and lose focus.

I remember reading a long time ago, in the LWotC issues section, about this problem; however, I can not find that post. Nonetheless, it was mentioned that it would likely be best to allow people to mod the recruit screen to suit their needs, just in case additional information needs to be added or removed from it.

To this effect, UIScreenListener_RecruitSoldiers has been commented out and I created, a number of months ago, a recruit screen mod that fixes these issues. The mod can be found at : https://steamcommunity.com/sharedfiles/filedetails/?id=2116413401.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIScreenListener_RecruitSoldiers.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIScreenListener_RecruitSoldiers.uc
@@ -1,12 +1,23 @@
 //---------------------------------------------------------------------------------------
 //  FILE:    UIScreenListener_RecruitSoldiers
 //  AUTHOR:  Peter Ledbrook
-//
 //  PURPOSE: A big ball of event listeners we need to set up for tactical games.
+//
+//	KDM : This recruit screen 'screen listener' has been disabled so that the recruit screen can be modded.
+//
+//	The former implementation was problematic for 2 reasons :
+//	1.] When you recruited a soldier from the recruit screen, the screen reverted to its base game 'look'.
+//	2.] Recruit attribute text did not change colour when different recruit rows were selected. 
+//	This is because, while the screen listener implemented OnReceiveFocus and OnLoseFocus, these were only 
+//	called when the screen itself received and lost focus; they were not called when individual list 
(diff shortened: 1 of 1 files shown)


pledbrook:
Can we not just integrate the changes from your mod, since it's basically fixing problems/bugs in LWOTC? Also, does your mod handle the situation where you have more recruits than fit on the screen and they can't be scrolled through?

kdm2k6:
We can integrate it as long as modding of the recruit screen, by other people, doesn't matter, since it uses a screen listener to pop off the old recruit screen and push on the custom one.

I tested it with a controller as well as mouse/keyboard and it scrolls through the soldiers, even if there are a lot of them. Was this a problem before? If you have a save where this happens please send it to me and I'll try it out.

pledbrook:
With the UI, we often have to override the base game stuff. I think listeners can be used to override another mod's overrides (as opposed to MCOs, which can't). Here's a save file replicating too many recruits in the Recruit screen:

[save_BUG Too many recruits.zip](https://github.com/long-war-2/lwotc/files/5610854/save_BUG.Too.many.recruits.zip)

I can't scroll through with the mouse wheel and the list of items don't fit into the parent screen.

kdm2k6:
Thanks. I'll give the save a try.

kdm2k6:
Just tested it with a mouse and keyboard and things appear to work with my mod enabled; I noticed the issue when it wasn't enabled as well.

pledbrook:
What's the plan? Will you update this PR with the mod's code? Would you like me to grab the mod's code and integrate it directly? Something else?

kdm2k6:
Already done! Just going to check it today then submit a new pull request.

PR #1178, opened 2020-11-30
Rework some of the Reaper ability tree

pledbrook:
- Blood Trail now applies a Dodge reduction bonus to the target instead of bleeding.
 - Soul Harvest grants +10% base crit chance plus +4% per kill up to an overall maximum of +30%.
 - Lone Wolf now has a tapering effect so that it doesn't lose the entire bonus if you're just within the minimum 7-tile range.
 - Shadow Grenadier is a new passive ability that allows the Reaper to throw grenades without breaking Shadow and grants a grenade pocket.
 - Poisoned Blades is a new passive ability that makes throwing knife attacks apply poison.
 - Charge Battery is a new charge-based ability that allows the Reaper to extend the duration of Shadow by 1 turn.
 - Knife Faceoff is a new ability with a single charge that behaves like Faceoff for throwing knives.
 - Needle, Remote Start, and Homing Mine have been moved to the random ability decks.


LongWarOfTheChosen/Config/XComLW_FactionBalance.ini
```diff
@@ -86,11 +89,19 @@ OVERRIDE_REDUCTION=10
 
 [LW_FactionBalance.X2LWModTemplate_ReaperAbilities]
 SHADOW_DURATION=1
-SHADOW_DETECTION_RANGE_REDUCTION=0.7f
+SHADOW_DETECTION_RANGE_REDUCTION=0.9f
```
Grobobobo, line 92: at this point why not go for 100%?

LongWarOfTheChosen/Config/XComLW_FactionBalance.ini
```diff
@@ -86,11 +89,19 @@ OVERRIDE_REDUCTION=10
...
 REMOTE_START_CHARGES=1
 REMOTE_START_DEMOLITIONIST_CHARGES=1
 
+BLOOD_TRAIL_ANTIDODGE_BONUS=20
+
+PALE_HORSE_BASE_CRIT=10
+PALE_HORSE_PER_KILL_CRIT=4
+PALE_HORSE_MAX_CRIT=20  ; Max crit from the per-kill bonus (does not include base crit bonus)
+
+STING_RUPTURE=1  ; Rupture amount inflicted by Sting
```
Grobobobo, line 103: Rupture does not stack, so 1 is on the low side, and becomes worthless on plasma

LongWarOfTheChosen/Src/LW_FactionBalance/Classes/X2Effect_PaleHorse_LW.uc
```diff
@@ -0,0 +1,29 @@
+//---------------------------------------------------------------------------------------
+//  FILE:    X2Effect_PaleHorse_LW
+//  AUTHOR:  Peter Ledbrook
+//  PURPOSE: Modified version of Pale Horse (Soul Harvest) that can apply to all
+//           weapons and has a base crit bonus in addition to the per-kill bonus.
```
Grobobobo, line 5: Not sure if this is the way to fix soul harvest, but it's hard to think of an interesting ramp up ability that isn't broken. maybe something like XCOM blood thirst, where the crit bonus is a lot bigger but only temporary?

LongWarOfTheChosen/Config/XComLW_FactionBalance.ini
```diff
@@ -86,11 +89,19 @@ OVERRIDE_REDUCTION=10
 
 [LW_FactionBalance.X2LWModTemplate_ReaperAbilities]
 SHADOW_DURATION=1
-SHADOW_DETECTION_RANGE_REDUCTION=0.7f
+SHADOW_DETECTION_RANGE_REDUCTION=0.9f
 
 REMOTE_START_CHARGES=1
 REMOTE_START_DEMOLITIONIST_CHARGES=1
 
+BLOOD_TRAIL_ANTIDODGE_BONUS=20
```
Grobobobo, line 97: Since the dodge malus here can't go below 0 to boost crit, I wonder if it'd be better to make the malus bigger, after all it does not affect graze band and mathematicaly 20 dodge is 10% dmg reduction. I'd try 30 or even 40 to make it feel more impactful even if most targets don't even have that much dodge.

Grobobobo (commented):
I think the current reaper could use the skirmisher treatment where we overbuff it to make people better used to it and then bring it in line with everything else, so i think some numbers could use a boost or two

LongWarOfTheChosen/Config/XComLW_FactionBalance.ini
pledbrook: At first, it seemed too cheesy. But then so is 0.9. Now I'm wondering whether it might break things as an edge case. We could try 1.0 and hope for the best. It might also be good to show concealment tiles as an additional indicator that the Reaper is in Shadow.

LongWarOfTheChosen/Config/XComLW_FactionBalance.ini
pledbrook: Are we thinking this is OK now? I was thinking that if you're reduced Rupture ability to 3, then 2 on Sting would be overly strong considering the holotargeting effect.

LongWarOfTheChosen/Src/LW_FactionBalance/Classes/X2Effect_PaleHorse_LW.uc
pledbrook: Maybe, but I think Soul Harvest is fine as it is now, especially as it applies to knives. I have to admit apathy on changing it further at this point. A temporary effect could be more interesting, but maybe try that at a later date.

LongWarOfTheChosen/Config/XComLW_FactionBalance.ini
pledbrook: Sure. Let's go ham and change it to 40.

LongWarOfTheChosen/Config/XComLW_FactionBalance.ini
Grobobobo: since apparently it DOES stack with other sources, rupture 1 is fine.

LongWarOfTheChosen/Config/XComLW_FactionBalance.ini
Grobobobo: I think the big yellow tint is indication enough, I'd go for 100%.

pledbrook: OK, I confirmed that the HUD is still there, so yes, concealment tiles aren't needed to indicate Shadow is active. OK, I'll try 1.0f.

PR #1180, opened 2020-12-01
Fix Recruitment Screen

kdm2k6:
The original Recruit Screen implementation was problematic for several reasons :

1.] When you recruited a soldier from the recruit screen, the screen reverted to its base game 'look'.
2.] Recruit attribute text did not change colour when different recruit rows were selected. This is because, while the screen listener implemented OnReceiveFocus and OnLoseFocus, these were only called when the screen itself received and lost focus; they were not called when individual list items received and lost focus.
3.] When there were many recruits available, the list fell off the recruit panel and wasn't scrollable.

----------------------

Modifies : UIScreenListener_RecruitSoldiers.uc

Makes sure the new Long War of the Chosen recruit screen, UIRecruitSoldiers_LWOTC, is pushed onto the screen stack instead of the base game's recruit screen.

----------------------

Added : UIRecruitSoldiers_LWOTC.uc

A recruitment screen which makes sure Long War of the Chosen recruitment list items, UIRecruitmentListItem_LWOTC, are pushed onto the recruitment list.

----------------------

Added : UIRecruitmentListItem_LWOTC.uc
Commented out code from : UIRecruitmentListItem_LW.uc

UIRecruitmentListItem_LWOTC is a slightly modified version of UIRecruitmentListItem_LW; it makes sure that stat text, within list items, changes colour based upon focus status.

----------------------

Added : X2EventListener_UpdateResources_RecruitScreen.uc

Makes sure that the resource bar, located in the top right corner of the screen, shows supplies and monthly supplies when the new recruitment screen is being displayed.

----------------------

Commented out code from : XComGameState_LWToolboxOptions.uc

Several event related functions are no longer needed with the new recruitment implementation.

----------------------

Modified : XComLW_UI.ini

Added a few config variables which determine recruit stat text size and location.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIRecruitSoldiers_LWOTC.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIRecruitSoldiers_LWOTC.uc
@@ -1,0 +1,46 @@
+//---------------------------------------------------------------------------------------
+//  FILE:	 UIRecruitSoldiers_LWOTC.uc
+//  AUTHOR:	 KDM
+//  PURPOSE: Long War of the Chosen compatible recruit screen.
+//--------------------------------------------------------------------------------------- 
+
+class UIRecruitSoldiers_LWOTC extends UIRecruitSoldiers;
+
+simulated function UpdateData()
+{
+	local int i;
+	local XComGameState_Unit Recruit;
+	local XComGameStateHistory History;
+	local XComGameState_HeadquartersResistance ResistanceHQ;
+
+	AS_SetTitle(m_strListTitle);
+
+	List.ClearItems();
+	m_arrRecruits.Length = 0;
+
+	History = `XCOMHISTORY;
+	ResistanceHQ = XComGameState_HeadquartersResistance(History.GetSingleGameStateObjectForClass(class'XComGameState_HeadquartersResistance'));
+
+	if (ResistanceHQ != none)
+	{
+		for (i = 0; i < ResistanceHQ.Recruits.Length; i++)
+		{
+			Recruit = XComGameState_Unit(`XCOMHISTORY.GetGameStateForObjectID(ResistanceHQ.Recruits[i].ObjectID));
+			m_arrRecruits.AddItem(Recruit);
+			// KDM : Create and add our custom list item to the list.
+			UIRecruitmentListItem_LWOTC(List.CreateItem(class'UIRecruitmentListItem_LWOTC')).InitRecruitItem(Recruit);
+		}
+	}
+
+	if (m_arrRecruits.Length > 0)
+	{
+		// KDM : If we don't call RealizeList(), the bottom of the list gets cut off after recruiting a soldier.
+		List.RealizeList();
+		List.SetSelectedIndex(0, true);
+	}
+	else
+	{
+		List.SetSelectedIndex(-1, true);
+		AS_SetEmpty(m_strNoRecruits);
+	}
+}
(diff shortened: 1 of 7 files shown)


LongWarOfTheChosen/Src/LW_Toolbox_Integrated/Classes/XComGameState_LWToolboxOptions.uc
```diff
@@ -522,11 +522,13 @@ simulated function RegisterListeners()
 	// angle rotation
 	EventManager.RegisterForEvent(ThisObj, 'OverrideCameraRotationAngle', OnGetCameraRotationAngle, ELD_Immediate,,,true);
 
-	// recruit listitem modification
-	EventManager.RegisterForEvent(ThisObj, 'OnRecruitmentListItemInit', AddRecruitStats, ELD_Immediate,,,true);
-	EventManager.RegisterForEvent(ThisObj, 'OnRecruitmentListItemUpdateFocus', UpdateRecruitFocus, ELD_Immediate,,,true);
+	// Recruit list item modification
+	// KDM : These events are no longer needed with our Long War of the Chosen compatible recruit screen, UIRecruitSoldiers_LWOTC.
+	// EventManager.RegisterForEvent(ThisObj, 'OnRecruitmentListItemInit', AddRecruitStats, ELD_Immediate,,,true);
+	// EventManager.RegisterForEvent(ThisObj, 'OnRecruitmentListItemUpdateFocus', UpdateRecruitFocus, ELD_Immediate,,,true);
```
pledbrook, line 528: As with earlier pull request, please just delete the code.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2EventListener_UpdateResources_RecruitScreen.uc
```diff
@@ -0,0 +1,51 @@
...
+}
+
+static function EventListenerReturn OnUpdateResources_RecruitScreen(Object EventData, Object EventSource, XComGameState GameState, Name Event, Object CallbackData)
+{
+	local XComHQPresentationLayer HQPres;
+
+	HQPres = `HQPRES;
+
+	// KDM : We are only interested when our custom Recruit Screen is on the top of the screen stack.
+	if (HQPres.ScreenStack.GetCurrentClass() == class'UIRecruitSoldiers_LWOTC')
```
pledbrook, line 42: What if another mod overrides our Recruit screen? I'm thinking it may be more friendly to check for subclasses here too.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIRecruitSoldiers_LWOTC.uc
```diff
@@ -0,0 +1,46 @@
+//---------------------------------------------------------------------------------------
+//  FILE:	 UIRecruitSoldiers_LWOTC.uc
+//  AUTHOR:	 KDM
+//  PURPOSE: Long War of the Chosen compatible recruit screen.
+//--------------------------------------------------------------------------------------- 
+
+class UIRecruitSoldiers_LWOTC extends UIRecruitSoldiers;
```
pledbrook, line 7: A `_LW` suffix is sufficient and preferred. I'm not that interested in identifiying classes that were added in LWOTC rather than LW2. The suffix is there just to identify vanilla classes that have been Long War-ized. Same goes for the other classes that currently have a `_LWOTC` suffix.

pledbrook (changes requested):
Thanks for this. On top of the requests inline, could you please add the new files to the *LongWarOfTheChosen.x2proj* file?

I'm also wondering whether it's better to use an MCO for the recruitment screen rather than the UIScreenListener. The MCO has the advantage that we shouldn't need to include the event listener that updates the resource numbers. Is there an advantage to the UIScreenListener? The only one I can think of is that it would allow another mod to MCO *our* recruit screen.

PR #1182, opened 2020-12-03
Personnel Screen Scrolling Fix :

kdm2k6:
Modifies : X2EventListener_Headquarters
Adds a listener function, OnUIPersonnelDataRefreshed, which listens for the event 'UIPersonnel_OnSortFinished', triggered within UIPersonnel.RefreshData.

The UIPersonnel screen scrolling problem can be seen most clearly via the "View Soldiers" screen while using a controller. If you enter the "View Soldiers" screen, click on a soldier to view their information, then exit back to the "View Soldiers" screen, you may notice that the soldier list item selection remains consistent. In other words, if you click on the list item corresponding to "Doug Smith", "Doug Smith" will still be selected when you return to the "View Soldiers" screen, after viewing his information/customization options/weapon etc.

The problem is that while list item selection remains consistent, the list scroll, required if the number of list items exceeds the height of the list, does not. So when you return to the "View Soldiers" screen, after viewing "Doug Smith's" information, you may not be able to see him any more, despite him still being selected. This is particularly evident in situations when you have, say, a list of 50 or 60 soldiers, and you choose the bottom-most soldier.

The underlying issue is that UIPersonnel's UpdateList function clears its list, thus removing and resetting its scrollbar, but never sets the scrollbar back to the appropriate location. To be more specific, as mentioned in my comments :

"UIPersonnel's UpdateList() calls SetSelectedIndex() on the list, and SetSelected() on the list's navigator; however, neither function modifies the list's scroll position via Scrollbar.SetThumbAtPercent(). This is because the only place the scrollbar is manipulated is within UIList's NavigatorSelectionChanged, which is called via Navigator.OnSelectedIndexChanged. Navigator.OnSelectedIndexChanged is called in functions like  Prev(), Next(), SelectFirstAvailable(), and SelectFirstAvailableIfNoCurrentSelection(); however, it is not called within SetSelected()."

This fix makes sure that the scrollbar, if one exists, is set correctly whenever RefreshData(), and consequently, UpdateList(), are called. This way we can always 'see' the selected list item.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2EventListener_Headquarters.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2EventListener_Headquarters.uc
@@ -37,6 +37,7 @@ static function CHEventListenerTemplate CreateXComHQListeners()
 	Template.AddCHEvent('OverrideScienceScore', OverrideScienceScore, ELD_Immediate, GetListenerPriority());
 	Template.AddCHEvent('CanTechBeInspired', CanTechBeInspired, ELD_Immediate, GetListenerPriority());
 	Template.AddCHEvent('UIAvengerShortcuts_ShowCQResistanceOrders', ShowOrHideResistanceOrdersButton, ELD_Immediate, GetListenerPriority());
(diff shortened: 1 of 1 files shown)


pledbrook (approved):
Great, thanks (again)!

PR #1184, opened 2020-12-04
Fix Recruit Screen using mod class override

kdm2k6:
The original Recruit Screen implementation was problematic for several reasons :

1.] When you recruited a soldier from the recruit screen, the screen reverted to its base game 'look'.
2.] Recruit attribute text did not change colour when different recruit rows were selected. This is because, while the screen listener implemented OnReceiveFocus and OnLoseFocus, these were only called when the screen itself received and lost focus; they were not called when individual list items received and lost focus.
3.] When there were many recruits available, the list fell off the recruit panel and wasn't scrollable.

--------------------------------

Modifies : XComEngine.ini

Adds a mod class override, substituting "LW_Overhaul.UIRecruitSoldiers_LW" for the base class "UIRecruitSoldiers".

--------------------------------

Removes : UIScreenListener_RecruitSoldiers.uc

We are using a mod class override so the screen listener is no longer needed.

--------------------------------

Adds : UIRecruitSoldiers_LW.uc

A recruitment screen which makes sure Long War of the Chosen recruitment list items, UIRecruitmentListItem_LW, are pushed onto the recruitment list.

--------------------------------

Moves : UIRecruitmentListItem_LW.uc from LW_Toolbox_Integrated to LW_Overhaul, and modifies UIRecruitmentListItem_LW.uc

1.] The file was moved to LW_Overhaul as this seems like the logical place to have it.
2.] The new version of UIRecruitmentListItem_LW is a slightly modified version which makes sure that stat text, within list items, changes colour based upon focus status.

--------------------------------

Modifies : X2EventListener_Headquarters.uc

Adds the listener function, OnUpdateResources_LW, which listens for the UIAvengerHUD event, UpdateResources. In this circumstance, it makes sure that the resource bar, located in the top right corner of the screen, shows supplies and monthly supplies when the new recruitment screen is being displayed.

--------------------------------

Modifies : XComGameState_LWToolboxOptions.uc

Removes the event listener functions AddRecruitStats, and UpdateRecruitFocus, along with event registration for OnRecruitmentListItemInit, and OnRecruitmentListItemUpdateFocus. These are no longer needed with the new recruit screen.

--------------------------------

Modifies : XComLW_UI.ini

Added a few config variables which determine recruit stat text size and location.


kdm2k6:
The resource display still needed to be updated via a listener since UIAvengerHUD only looks at UIRecruitSoldiers, and I don't believe even a class override gets around this. Additionally, I tested it, and nothing showed up before adding the listener code.

PR #1185, opened 2020-12-04
Capital letters fix

FlashVanMaster:
I just realize i forgot that stats should be with  capital letters (also some changes to basic lw2)
also some words fixes


pledbrook:
Merged

PR #1187, opened 2020-12-07
Fix Dark Event Background and Resource Bar

kdm2k6:
Problem 1 : The Dark Event background does not appear.
Modifies : UIAdventOperations_LW.BuildScreen

The Long War 2 Dark Events screen, UIAdventOperations_LW, is derived from the class UIAdventOperations; this class makes use of the graphics/flash package : gfxDarkEvents.upk. When looking at the flash code, it becomes apparent that 2 graphical elements, bottomShadow and hexBackground, are only made visible through the function AnimateIn(). Unfortunately, Long War had originally commented out this code, which was called within the UnrealScript function BuildScreen. Additionally, I subsequently removed this commented out code. Therefore, these flash elements were no longer being shown.

Adding the call to AnimateIn(), via MC.FunctionVoid("AnimateIn"), fixes this issue.

Problem 2 : The Dark Event resource bar doesn't appear
Modifies : X2EventListener_Headquarters.OnUpdateResources_LW

This is a common problem in various areas. UIAvengerHUD.UpdateResources shows the resource bar, with the appropriate resources, according to the screen being shown; unfortunately, it does not take subclasses into account. In this particular case, it looks for UIAdventOperations, but not the Long War subclass UIAdventOperations_LW. Consequently, an event listener needed to be set up to show the monthly supplies, supplies, and intel, just like in the base game.

Fixes X2EventListener_Headquarters.OnUpdateResources_LW logic

X2EventListener_Headquarters.OnUpdateResources_LW has been modified such that its code only runs if 'subclasses' are on top of the screen stack. This is because if, for some reason, this code is run on the base Recruit Screen, or base Dark Event Screen, we will see double resources since UIAvengerHUD.UpdateResources has already dealt with the resource bar.

Modifies : UIAdventOperations_LW.OnRevealClicked

When a dark event is revealed, intel has been used to do so; consequently, a call to UIAvegerHUD's UpdateResources is now made. This updates the resource bar which is now visible on the dark event screen.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIAdventOperations_LW.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIAdventOperations_LW.uc
@@ -49,6 +49,14 @@ simulated function BuildScreen( optional bool bAnimateIn = false )
 	// many active or pending dark events there are.
 	MC.FunctionNum("SetNumCards", 3);
 	RefreshNav();
+
+	// KDM : Previously, the flash call to AnimateIn had been commented out by Long War; however,
+	// this created a situation in which 2 flash elements, bottomShadow and hexBackground, were never
+	// made visible on the Dark Events screen.
+	if (bAnimateIn)
+	{ 
+		MC.FunctionVoid("AnimateIn");
+	}
 }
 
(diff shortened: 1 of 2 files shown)


pledbrook (approved):
Thanks

PR #1191, opened 2020-12-11
Clean up UIPersonnel_SoldierListItem_LW Focus and Disabled Code

kdm2k6:
UIPersonnel_SoldierListItem_LW is a UIButton, and ultimately a UIPanel.

UIPanel's bIsFocused is updated whenever a UIPanel recieves focus, via OnReceiveFocus, or loses focus, via OnLoseFocus. Consequently :
1.] UIPersonnel_SoldierListItem_LW.UpdateItemsForFocus's parameter, Focussed, is unnecessary.
2.] The boolean variable bIsFocussed, within UIPersonnel_SoldierListItem_LW, is unnecessary.

Just as a reminder, within OnReceiveFocus and OnLoseFocus, it is important to call their super variant, super.OnReceiveFocus or super.OnLoseFocus, before UpdateItemsForFocus. This is because UpdateItemsForFocus makes use of a UIPanel's focus status, defined by bIsFocused and updated via UIPanel.OnReceiveFocus and UIPanel.OnLoseFocus. Originally, within LW2, UpdateItemsForFocus was passed a true or false value to represent focus state; hence, function call order was unimportant.

UIButton's IsDisabled is updated whenever UIButton.SetDisabled is called. Consequently :
1.] To simplify code, the Disabled parameter from GetWillValueColor, and GetCombatIntelligenceColor, has been removed; instead IsDisabled is referenced within these functions. This does not change any functionality, since IsDisabled was always sent in as the parameter value anyways.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIPersonnel_SoldierListItem_LW.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIPersonnel_SoldierListItem_LW.uc
@@ -9,8 +9,6 @@ class UIPersonnel_SoldierListItem_LW extends UIPersonnel_SoldierListItem config(
 var float IconXPos, IconYPos, IconXDelta, IconScale, IconToValueOffsetX, IconToValueOffsetY, IconXDeltaSmallValue;
 var float DisabledAlpha;
 
-var bool bIsFocussed;
-
 // Icons to be shown in the class area
 var UIImage AimIcon, WillIcon;
 var UIText AimValue, WillValue;
@@ -28,10 +26,7 @@ simulated function UIButton SetDisabled(bool disabled, optional string TooltipTe
 {
 	super.SetDisabled(disabled, TooltipText);
 	UpdateDisabled();
-	// KDM : Originally, UpdateItemsForFocus was called with the parameter 'false'; however, this was problematic
-	// since ability text colour would no longer depend upon the list item's focus state; it would always 'assume'
-	// it was unfocused. In order to take the list item's focus into account, we need to send in bIsFocused.
-	UpdateItemsForFocus(bIsFocused);
+	UpdateItemsForFocus();
 	return self;
 }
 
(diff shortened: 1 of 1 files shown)


LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIPersonnel_SoldierListItem_LW.uc
```diff
@@ -28,10 +26,7 @@ simulated function UIButton SetDisabled(bool disabled, optional string TooltipTe
 {
 	super.SetDisabled(disabled, TooltipText);
 	UpdateDisabled();
-	// KDM : Originally, UpdateItemsForFocus was called with the parameter 'false'; however, this was problematic
-	// since ability text colour would no longer depend upon the list item's focus state; it would always 'assume'
-	// it was unfocused. In order to take the list item's focus into account, we need to send in bIsFocused.
-	UpdateItemsForFocus(bIsFocused);
+	UpdateItemsForFocus();
```
pledbrook, line 29: In the commit message you say it's important to call `OnReceiveFocus()` or `OnLoseFocus()` before `UpdateItemsForFocus()`, but that doesn't happen here. Some clarification would be helpful.

kdm2k6: I've updated the commit message and message here.

`Within OnReceiveFocus and OnLoseFocus, it is important to call their super variant, super.OnReceiveFocus or super.OnLoseFocus, before UpdateItemsForFocus. This is because UpdateItemsForFocus makes use of a UIPanel's focus status, defined by bIsFocused and updated via UIPanel.OnReceiveFocus and UIPanel.OnLoseFocus. Originally, within LW2, UpdateItemsForFocus was passed a true or false value to represent focus state; hence, function call order was unimportant.`

`SetDisabled` is not concerned with "setting" focus status; therefore, it need not deal with `OnReceiveFocus` nor `OnLoseFocus`. It will take whatever the list item's focus state is, represented through `bIsFocused`, and update accordingly.

pledbrook (approved):
Thanks, looks good

PR #1194, opened 2020-12-13
Allow Contacted Haven Cycling with Controller

kdm2k6:
Modifies : UIScreenListener_UIStrategyMap.uc

Controller users can now cycle through 'contacted' havens/regions on the strategy map using DPad left and DPad right. The following rules are followed :
1.] If a 'contacted' region is currently not selected, then a DPad press will select the starting region.
2.] If a 'contacted' region is currently selected, then DPad left will cycle 'contacted' regions from right to left. Although there is no way to define which regions are to the left/right of others, since this is based upon XComGameStateHistory's storage of outposts, selection order always remains consistent. During testing, this seemed to be a completely acceptable limitation, as map items can be cycled through quite quickly.
3.] If a 'contacted' region is currently selected, then DPad right will cycle 'contacted' regions from left to right.

Modifies XComLW_UI.ini

Adds a config variable, CYCLE_HAVENS_INSTANTLY, which allows you to determine if haven cycling should be instantaneous or with acceleration/deceleration. The latter is the default option, as instantaneous movement, while much faster, can look a tad jarring.

Fix in : UIScreenListener_UIStrategyMap.uc

All controller button presses are now ignored if the Avenger is in flight mode; this is consistant with default XCom 2 code.

### DIFF
--- a/LongWarOfTheChosen/Config/XComLW_UI.ini
+++ b/LongWarOfTheChosen/Config/XComLW_UI.ini
@@ -79,3 +79,6 @@ RECRUIT_FONT_SIZE_CTRL = 22					; Recruit screen font size for controller users.
 RECRUIT_Y_OFFSET_CTRL = -3					; Moves the stats up/down; negative numbers move up / positive numbers move down.
 RECRUIT_FONT_SIZE_MK = 20					; Recruit screen font size for mouse & keyboard users.
 RECRUIT_Y_OFFSET_MK = -2;					; Moves the stats up/down; negative numbers move up / positive numbers move down.
+
+[LW_Overhaul.UIScreenListener_UIStrategyMap]
+CYCLE_HAVENS_INSTANTLY = false				; When controller users cycle 'contacted' havens with the DPad, do we want the strategy map movement to be instantaneous or with acceleration/deceleration ?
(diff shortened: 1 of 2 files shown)


pledbrook (commented):
This is fantastic, thanks! One thing: I'm not seeing a nav helper for opening the haven management screen. I eventually found out through experimentation that it was X, but I think that should be made clear if possible.

kdm2k6:
Correct. I was aware of this when I first dealt with Resistance/Haven screens; however, it needs a highlander hook.

If you create a highlander hook at the bottom of `UIStrategyMap.UpdateButtonHelp` I will add help for opening up the Resistance menu via X as well as cycling havens via the DPad.

**One note** : The X button opens up the Resistance menu if no haven is selected; however, the X button opens up the selected haven menu if one is selected. Despite this, I will make a simple navigation tip; something along the lines of "Open Resistance Menu / Haven Menu". I could theoretically, via a hook within UIStrategyMap.SelectMapItemNearestLocation have the tip change based on whether a haven is selected or not. The **problem** is that this results in a lot of navigation help flickering. XCom 2 has a strange navigation help system; you can clear it and add to it, but you can't go in and add/modify/delete elements. Therefore, every time you want to refresh it you have to clear it then create it anew - this creates flickering which I do not like to consistently happen, as it is quite distracting.

pledbrook:
Could you please raise a highlander issue detailing the request? So it doesn't get lost.

kdm2k6:
https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/issues/932.
Once it is done, I'll add the navigation help.

pledbrook (approved):
Looks good, thanks

PR #1197, opened 2020-12-17
Alien Overwatch AI

NilchEi:
* Removed ChosenSafeToMove from TryPartingSilk as Grobobobo suggests, as testing revealed that overwatches had significant chance of preventing guaranteed melee attack. Also removed outdated and unnecessary DisableExtractThisTurn child.

* Allowed ChosenSafeToMove to recognize LightningReflexes_LW as well, instead of only Shadowstep.

* EvacWithinVisRange added to AvoidDoubleMove.
When trying to sneak a unit to evac near activated enemies, completely breaking line of sight is extremely effective as no enemy visibility makes ConsiderTakingOverwatchEvac check fail; all enemies will choose to double move instead of overwatching near evac.
Added EvacWithinVisRange check to LW2 AvoidDoubleMove behavior, so enemy near evac knows to overwatch with their last action to prevent escape, even when they do not have enemy visibility on their current position.
AvoidDoubleMove is last child called by TryOverwatchLastAction, which is the last child of TryShootOrReloadOrOverwatch. This therefore have no impact during active engagement and only serves as counter to stealth strategy.


pledbrook:
Merged with [this commit](https://github.com/long-war-2/lwotc/commit/f6206a03a9ab5adc658542c2eb91c47967cb35c9)

PR #1199, opened 2020-12-18
Add compatibility with extra inventory slots

K1ruka:
This will add compatibility for Spark Arsenal's Auxiliary Slot, Spark Grenade Slot, and BIT Heavy Weapon Slot, as well as Psionics Overhaul V3's Psi Amp Slot, so that none of those equipment slots can be messed with on a soldier that's infiltrating. I also removed the duplicate GrenadePocket line because it doesn't really need to be there.


pledbrook:
I implemented this as a configurable array in [this commit](https://github.com/long-war-2/lwotc/commit/5b03ddebbb52bb5f1d89862dbd870ef06b07a729).

PR #1201, opened 2020-12-19
Fix Controller Tactical Ability Selection + Add Commander Range function

kdm2k6:
Modifies :  UITacticalHUD_AbilityContainer_LWExtended

1.] Adds : ChangeAbilityPage
Formerly, both OnScrollButtonClicked and OnUnrealCommand contained code capable of changing the ability page; consequently, there was a great deal of code duplication. This code has been consolidated into the new function ChangeAbilityPage; this function allows the ability page to be moved backwards or forwards, depending upon the parameter PreviousPage.

2.] Modifies : UpdateAbilitiesArray
Added a bit of code, previously only found in UITacticalHUD_AbilityContainer.UpdateAbilitiesArray, which updates the visibility of the 'Call Skyranger' navigation help. As a result, the navigation help appears whenever Skyranger Evac is possible, although it does hide this information when the ability menu is raised so as not to get in the way of enemy targeting.

3.] Modifies : DirectConfirmAbility
After activating UITacticalHUD_MouseControls_LWOfficerPack for controller users, so they could access 'officer range' functionality, the 'Call Skyranger' ability stopped working. The reason is that DirectConfirmAbility, which the 'Call Skyranger' ability makes use of, prevents 'command' abilities from going through. This works well for mouse and keyboard users since it prevents them from hitting a number key and activating an ability which is not on the ability menu. Since controller users can not hit number keys, DirectConfirmAbility need not ignore 'command' abilities, such as 'Call Skyranger', when a controller is active.

4.] Modifies : CycleAbilitySelection
CycleAbilitySelection is concerned with changing ability selection when either DPad left or DPad right is pressed; formerly, it was not working properly since it did not 'understand' LW2's new 'single row / multi page' ability system. Now, the user can press DPad left / Dpad right to move between a single ability page's abilities; furthermore, pressing DPad left on the first ability of a page wraps around to the last ability on a page while pressing DPad right on the last ability of a page wraps around to the first ability on a page.

5.] Modifies : CycleAbilitySelectionRow
CycleAbilitySelectionRow was originally concerned with changing the ability row when either DPad up or DPad down were pressed; it was not working properly since it did not 'understand' LW2's new 'single row / multi page' ability system. Now, the user can press DPad up / Dpad down to move between ability pages; furthermore, pressing DPad up on the first ability page wraps around to the last ability page while pressing DPad down on the last ability page wraps around to the first ability page.

6.] Modifies : NotifyCanceled
Now, when controller users lower the ability menu, ability item focus is properly lost and the first ability page will be shown.

7.] Modifies : GetCurrentSelectedAbility
There was a rather complicated issue, described in comments above the function, resulting in a situation whereby enemy target heads were not updating whenever ability selection was being changed on the second/third/forth ... etc. ability page. Luckily the fix turned out to be relatively straightforward; it involved calling the base code's GetAbilityAtIndex function, rather than the LW2 variant of GetAbilityAtIndex.

8.] Modifies : OnAccept
Makes sure that selecting an inactive ability no longer results in it losing 'visual' focus. Formerly, if you selected an inactive ability, its selection brackets would disappear, even though it was still actually focused.

9.] Modifies : OnUnrealCommand
Controller users can no longer change ability pages with the X button; this was killing the reload X button and is no longer needed since DPad up / DPad down change ability pages when the ability menu is raised.

Modifies : UIScreenListener_TacticalHUD_LWOfficerPack

UITacticalHUD_MouseControls_LWOfficerPack is now spawned for controller users so they have access to the 'Officer range' functionality built into LW2.

Toggling the 'officer range' graphic, assuming an officer is on the mission, is a 2 step process for a controller user :
A.] Bring up the ability menu with the right trigger.
B.] Click on the left trigger.

I created a mod a while back for Long War Rebalanced which follows a similar principle, if a user brings up the ability menu they can then quickly reload, hunker down, or place a soldier in standby via a particular button press. This gets around the issue of limited controller buttons; furthermore, since left trigger is not used when the ability menu is up, it does not get in the way of any other functionality.  Finally, having used it extensively within LWR, I have found it extremely quick and convenient to use.

Modifies : UITacticalHUD_MouseControls_LWOfficerPack

1.] Modifies UpdateControls
If a controller is active, the normal mouse buttons in the top right corner of the screen are hidden since they interfere with the 'Call Skyranger' navigation help. Additionally, the positioning of the officer range icon has been modified a bit.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_OfficerPack_Integrated/Classes/UIScreenListener_TacticalHUD_LWOfficerPack.uc
+++ b/LongWarOfTheChosen/Src/LW_OfficerPack_Integrated/Classes/UIScreenListener_TacticalHUD_LWOfficerPack.uc
@@ -1,70 +1,98 @@
 //---------------------------------------------------------------------------------------
 //  FILE:    UIScreenListener_TacticalHUD_LWOfficerPack
 //  AUTHOR:  Amineri (Pavonis Interactive)
-//
 //  PURPOSE: Adds button to toggle officer command range preview
-//
 //--------------------------------------------------------------------------------------- 
 
 class UIScreenListener_TacticalHUD_LWOfficerPack extends UIScreenListener;
 
-// This event is triggered after a screen is initialized
 event OnInit(UIScreen Screen)
 {
 	local UITacticalHUD HUDScreen;
-
-	`log("Starting TacticalHUD Listener OnInit");
-
-	if(Screen == none)
-	{
-		`RedScreen("UITacticalHUD ScreenListener: Screen == none");
-		return;
-	}
+	
 	HUDScreen = UITacticalHUD(Screen);
-	if(HUDScreen == none)
+	
+	if (HUDScreen == none)
 	{
 		`RedScreen("UITacticalHUD ScreenListener: HUDScreen == none");
 		return;
 	}
+
 	UpdateMouseControls(HUDScreen);
-
-	//Not working because can't create new archetype in UnrealEd
-	// now spawn the CommandRange Actor, passing in the Archetype reference as the template
-	//CRActor = ParentScreen.Spawn(class'LWCommandRange_Actor', ParentScreen, 'CommandRange',,,CRTemplate);
-
 	class'LWOfficerUtilities'.static.GCandValidationChecks();
+	`SCREENSTACK.SubscribeToOnInputForScreen(Screen, OnUITacticalHUDCommand);
 }
 
-// This event is triggered after a screen receives focus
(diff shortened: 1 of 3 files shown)


pledbrook (approved):
This is too much for me to review in code I'm not familiar with, so I'm going to approve and leave it to the players to find any issues :smile:

Thanks!

kdm2k6:
Thanks. I know there is a lot there; however, I am fairly thorough when creating the code as well as checking it.
Worst case, I'll fix any small bugs which appear.

PR #1204, opened 2020-12-24
Allow Hiding of 'Save Squad' button + Squad Container

kdm2k6:
Provides 2 config variables which allow a controller user to hide the 1.] 'Save Squad' button and the 2.] Squad Container. These are turned OFF by default.

1.] The 'Save Squad' button appears while viewing a Long War squad's soldiers; controller users can not click it and, in fact, it appears not to 'do anything'.
2.] The 'Squad Container' allows mouse and keyboard users to perform squad selection on the Squad Select screen, as well as display the currently active squad. Once again, controller users can not click it, and my controller-capable Squad Management system - https://steamcommunity.com/sharedfiles/filedetails/?id=2314584410 - creates its own squad selection and display UI which overlaps with LW2's.

Both of these UI elements are created within UIScreenListener_SquadSelect_LW.OnInit; furthermore, the 'Squad Container' is actually initialized on a timed delay. Consequently, I am unable to hide these elements within my controller-capable Squad Management system without worrying about mod load order and timers. Therefore, the simplest course of action is to allow users to hide them via config variables.

### DIFF
--- a/LongWarOfTheChosen/Config/XComLW_UI.ini
+++ b/LongWarOfTheChosen/Config/XComLW_UI.ini
@@ -79,7 +79,3 @@ RECRUIT_FONT_SIZE_CTRL = 22					; Recruit screen font size for controller users.
 RECRUIT_Y_OFFSET_CTRL = -3					; Moves the stats up/down; negative numbers move up / positive numbers move down.
 RECRUIT_FONT_SIZE_MK = 20					; Recruit screen font size for mouse & keyboard users.
 RECRUIT_Y_OFFSET_MK = -2;					; Moves the stats up/down; negative numbers move up / positive numbers move down.
-
-[LW_Overhaul.UIScreenListener_SquadSelect_LW]
-HIDE_SAVESQUAD_BTN_WITH_CONTROLLER = false	; Hide the 'Save Squad' button which appears while viewing a Long War squad's soldiers.
-HIDE_SQUADCONTAINER_WITH_CONTROLLER = false	; Hide the 'Squad Container' which allows for squad selection on the Squad Select screen.
--- a/LongWarOfTheChosen/Config/XComLW_UI.ini
+++ b/LongWarOfTheChosen/Config/XComLW_UI.ini
@@ -79,3 +79,7 @@ RECRUIT_FONT_SIZE_CTRL = 22					; Recruit screen font size for controller users.
 RECRUIT_Y_OFFSET_CTRL = -3					; Moves the stats up/down; negative numbers move up / positive numbers move down.
 RECRUIT_FONT_SIZE_MK = 20					; Recruit screen font size for mouse & keyboard users.
 RECRUIT_Y_OFFSET_MK = -2;					; Moves the stats up/down; negative numbers move up / positive numbers move down.
+
+[LW_Overhaul.UIScreenListener_SquadSelect_LW]
+HIDE_SAVESQUAD_BTN_WITH_CONTROLLER = false	; Hide the 'Save Squad' button which appears while viewing a Long War squad's soldiers.
+HIDE_SQUADCONTAINER_WITH_CONTROLLER = false	; Hide the 'Squad Container' which allows for squad selection on the Squad Select screen.
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIScreenListener_SquadSelect_LW.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIScreenListener_SquadSelect_LW.uc
@@ -24,8 +24,6 @@ var GeneratedMissionData MissionData;
 var config float SquadInfo_DelayedInit;
 var const array<string> PlotTypes;
 
-var config(LW_UI) bool HIDE_SAVESQUAD_BTN_WITH_CONTROLLER, HIDE_SQUADCONTAINER_WITH_CONTROLLER;
-
 // This event is triggered after a screen is initialized
 event OnInit(UIScreen Screen)
 {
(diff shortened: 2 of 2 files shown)


pledbrook:
Not really a fan of this. The two options I'm happy with are:

1. If controller users can't actually interact with them, just hide them (not sure there's much point having the config var)
2. Integrate your mod :smile:

I absolutely would prefer the latter, since I would like LWOTC to be controller friendly out of the box. But I don't know how you feel about that. Thoughts?

A third option temporarily would be to only hide the buttons if your mod is loaded. There are a few such conditions dotted around the code. It's not great hard coding such exceptions, but it's generally a better user experience.

kdm2k6:
I would like to go with option 1 for now. 

Once I have finished off all of the essential controller-support stuff, so I am able to 'fully' play a campaign I will integrate it into LWotC.

Right now I am on a roll, so I want to keep on going; however, there isn't a ton left to do. Once I have nothing left to think about, I can really bear down on integration, which will take some time and thought.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIScreenListener_SquadSelect_LW.uc
```diff
@@ -100,10 +111,17 @@ event OnInit(UIScreen Screen)
...
 		}
-		// create the SquadContainer on a timer, to avoid creation issues that can arise when creating it immediately, when no pawn loading is present
-		SquadContainer = SquadSelect.Spawn(class'UISquadContainer', SquadSelect);
-		SquadContainer.CurrentSquadRef = SquadMgr.LaunchingMissionSquad;
-		SquadContainer.DelayedInit(default.SquadInfo_DelayedInit);
+
+		// KDM : Allow the 'Squad Container', which deals with squad selection on the Squad Select screen,
+		// to be hidden through non-creation. My controller-capable Squad Management mod has its own squad 
+		// selection and display UI which overlaps with LW2's UI.
+		if(!(`ISCONTROLLERACTIVE && HIDE_SQUADCONTAINER_WITH_CONTROLLER))
```
pledbrook, line 118:
```suggestion
		if (!`ISCONTROLLERACTIVE)
```

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIScreenListener_SquadSelect_LW.uc
```diff
@@ -65,9 +67,18 @@ event OnInit(UIScreen Screen)
 	bInSquadEdit = `SCREENSTACK.IsInStack(class'UIPersonnel_SquadBarracks');
 	if(bInSquadEdit)
 	{
-		SquadSelect.LaunchButton.OnClickedDelegate = OnSaveSquad;
-		SquadSelect.LaunchButton.SetText(strSquad);
-		SquadSelect.LaunchButton.SetTitle(strSave);
+		if(`ISCONTROLLERACTIVE && HIDE_SAVESQUAD_BTN_WITH_CONTROLLER)
```
pledbrook, line 70:
```suggestion
		if (`ISCONTROLLERACTIVE)
```

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIScreenListener_SquadSelect_LW.uc
```diff
@@ -24,6 +24,8 @@ var GeneratedMissionData MissionData;
 var config float SquadInfo_DelayedInit;
 var const array<string> PlotTypes;
 
+var config(LW_UI) bool HIDE_SAVESQUAD_BTN_WITH_CONTROLLER, HIDE_SQUADCONTAINER_WITH_CONTROLLER;
+
```
pledbrook, line 28:
```suggestion
```

LongWarOfTheChosen/Config/XComLW_UI.ini
```diff
@@ -79,3 +79,7 @@ RECRUIT_FONT_SIZE_CTRL = 22					; Recruit screen font size for controller users.
 RECRUIT_Y_OFFSET_CTRL = -3					; Moves the stats up/down; negative numbers move up / positive numbers move down.
 RECRUIT_FONT_SIZE_MK = 20					; Recruit screen font size for mouse & keyboard users.
 RECRUIT_Y_OFFSET_MK = -2;					; Moves the stats up/down; negative numbers move up / positive numbers move down.
+
+[LW_Overhaul.UIScreenListener_SquadSelect_LW]
+HIDE_SAVESQUAD_BTN_WITH_CONTROLLER = false	; Hide the 'Save Squad' button which appears while viewing a Long War squad's soldiers.
+HIDE_SQUADCONTAINER_WITH_CONTROLLER = false	; Hide the 'Squad Container' which allows for squad selection on the Squad Select screen.
```
pledbrook, line 85:
```suggestion
```

pledbrook (changes requested):
Let me know if you're OK with the proposed changes, or whether you have an argument for keeping the config vars.

LongWarOfTheChosen/Config/XComLW_UI.ini
kdm2k6: No problems with these changes - this works even better than config variables.
Also, like I said before, I'll look at integrating things in a little while when 'other controller tasks' die down.
Thanks.

kdm2k6:
Thought I replied to this earlier today; however, I can't find the reply.
In any case, the changes look good.

I'll look at integrating my mod in a little while, after I've completed other controller related tasks.

PR #1206, opened 2020-12-28
Fix the Objective Screen

kdm2k6:
LW2 attempted to add scrollbar support to objective descriptions through a screen listener which would create a custom `UITextContainer`; unfortunately, their method did not work and the objective description area could become an overlapping text mess.

Modifies : `UIScreenListener_ViewObjectives`

1.] For mouse and keyboard users :
- Scrollbars are enabled for the objective description text container via a Flash call to `EnableScrollbar`.
- Whenever an objective is chosen, via the objective list, scrolling is reset via a Flash call to `ResetScroll` and the scroll bar's slider position is reset via a Flash call to `SetThumbAtPercent`.
- A hidden panel is created and placed above the objective description's parent, `ItemCard`. Furthermore, it is set up to process the mouse events `FXS_MOUSE_SCROLL_DOWN` and `FXS_MOUSE_SCROLL_UP`. This allows users to scroll the objective description, via the mouse wheel, without having to be right 'on top' of the scroll bar.

2.] Controller users can now scroll the objective's description using the right stick up/down; these 'button' presses are caught and handled with the help of `SubscribeToOnInputForScreen`.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIScreenListener_ViewObjectives.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIScreenListener_ViewObjectives.uc
@@ -1,54 +1,174 @@
 //---------------------------------------------------------------------------------------
 //  FILE:    UIScreenListener_ViewObjectives.uc
-//  AUTHOR:  Amineri (Pavonis Interactive)
-//  PURPOSE: Adjust objectives pane to use scrollbar instead of slow autoscroll
+//  AUTHOR:  KDM
+//  PURPOSE: The objective screen modified such that objective descriptions are scrollable.
 //---------------------------------------------------------------------------------------
 
 class UIScreenListener_ViewObjectives extends UIScreenListener config(LW_Overhaul);
 
+var UIViewObjectives ObjectiveScreen;
+var UIPanel MouseEventPanel;
+
+var localized string ScrollObjectiveDescriptionStr;
+
 event OnInit(UIScreen Screen)
 {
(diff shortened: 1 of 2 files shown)


pledbrook:
Thank you , thank you, thank you! This has bugged me, but never been high enough priority to invest the time in fixing it.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIScreenListener_ViewObjectives.uc
```diff
@@ -1,54 +1,115 @@
...
 
 class UIScreenListener_ViewObjectives extends UIScreenListener config(LW_Overhaul);
 
+var UIViewObjectives ObjectiveScreen;
+var UIPanel MouseEventPanel;
+
 event OnInit(UIScreen Screen)
 {
-	MakeDescriptionUseScrollbar(UIViewObjectives(Screen));
+	// KDM : Controller users keep the auto-scroll behaviour built into the objective screen.
```
pledbrook, line 15: Isn't this kind of frustrating? I've noticed other games linking the scroll to a stick or the D-pad. Is that feasible?

kdm2k6: I'll try to take a look in the next few days and see how easy it is to do.

kdm2k6: Done. I did something similar in my Squad Management mod, so it was easier than I thought.

pledbrook (approved):
This is great, thanks

PR #1207, opened 2020-12-28
Make Resistance/Outpost Screen controls a bit more consistent.

kdm2k6:
When viewing the strategy map, the X button brings up the Resistance Management screen. Consequently, I think it is more consistent if, while on the Resistance Management screen, the X button brings up the selected haven rather than the A button. Furthermore, it feels better if pressing the X button on a Haven screen modifies the haven adviser, rather than through right stick click.

Modifies : UIResistanceManagement_LW.uc

On the Resistance Management screen the A button sends the Avenger to the selected haven; formerly, this was attached to right stick click.

Modifies : UIResistanceManagement_ListItem.uc

On the Resistance Management screen the X button opens the corresponding haven screen; formerly, this was attached to the A button.

Modifies : UIOutpostManagement.uc

When viewing a particular haven, the X button now [1] Opens the haven adviser selection screen, if no haven adviser currently exists. [2] Removes the haven adviser, if one currently exists. Formerly, this was attached to right stick click.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIResistanceManagement_LW.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIResistanceManagement_LW.uc
@@ -435,13 +435,13 @@ simulated function RefreshNavHelp()
 	// KDM : Controller specific navigation help
 	if (`ISCONTROLLERACTIVE)
 	{
-		// KDM : A button opens the Haven screen
-		NavHelp.AddLeftHelp(ViewHavenStr, class'UIUtilities_Input'.static.GetAdvanceButtonIcon());
+		// KDM : X button opens the Haven screen
+		NavHelp.AddLeftHelp(ViewHavenStr, class'UIUtilities_Input'.static.GetGamepadIconPrefix() $ class'UIUtilities_Input'.const.ICON_X_SQUARE);
 
-		// KDM : Right stick click sends the Avenger to the selected haven, if on the strategy map.
+		// KDM : A button sends the Avenger to the selected haven, if on the strategy map.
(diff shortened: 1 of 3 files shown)


pledbrook:
Would this one benefit from the highlander event I added for strategy map nav help?

kdm2k6:
Not in this case. The Resistance Management and Haven Management screens, as well as their navigation help, are already working well.

This is just changing a couple buttons around, in terms of functionality and corresponding navigation help symbols. For example, when viewing the Resistance Management screen, you would now press the X button to view the selected Haven rather than the A button.

The other request needed a highlander change since it was dealing with UIStrategyMap; these are dealing with LW classes which I have 'access' to.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIResistanceManagement_LW.uc
```diff
@@ -435,13 +435,13 @@ simulated function RefreshNavHelp()
 	// KDM : Controller specific navigation help
 	if (`ISCONTROLLERACTIVE)
 	{
-		// KDM : A button opens the Haven screen
-		NavHelp.AddLeftHelp(ViewHavenStr, class'UIUtilities_Input'.static.GetAdvanceButtonIcon());
-
-		// KDM : Right stick click sends the Avenger to the selected haven, if on the strategy map.
+		// KDM : X button opens the Haven screen
+		NavHelp.AddLeftHelp(ViewHavenStr, class'UIUtilities_Input'.static.GetGamepadIconPrefix() $ class'UIUtilities_Input'.const.ICON_X_SQUARE);
+
```
pledbrook, line 440:
```suggestion

```

pledbrook (approved):
Looks good, thanks

PR #1211, opened 2021-01-04
Provide shorter Promotion screen option without panel tearing

kdm2k6:
NOTE : Whenever I refer to a promotion screen, I am referring to the "New Promotion Screen by Default" promotion screen.

**Problem** : If you enter a soldier's promotion screen then quickly cycle through soldiers, using a controller or a mouse and keyboard, you will notice strange panel tearing on the right side of the screen.

**Explanation** : It turns out that when you choose to view a soldier's promotions, you are actually viewing a 3D Flash movie which is embedded with : [1] a `NPSBDP_UIArmory_PromotionHero` promotion screen [2] a bunch of other stuff, including 3D models. Interestingly, it appears that the promotion screen is given a certain area of the movie with which to occupy; a viewport of sorts. Furthermore, the promotion screen appears incapable of moving outside of this bounding rectangle.

Now, if a soldier has more than 7 ranks, `NPSBDP_UIArmory_PromotionHero.OnInit` calls the function `ResizeScreenForBrigadierRank` whose goal is to add space for an 8th rank column. Within this function :

1. The promotion screen width is increased
2. A variety of background panel's have their width increased to fit the new 8th column.
3. The promotion screen is shifted to the right so that the 8th column is placed in the correct position.
4. Many of the panels, which are now too far to the right due to step #3, are moved back to the left a bit.

The main problem here is that the promotion screen width increase, combined with the UI element shifting, results in the main background panel being placed upon against the bounding rectangle given to the promotion screen by the 3D Flash movie. This, by itself, isn't a problem. However, when a distortion effect is added in, you end up with parts of the background panel moving outside of the promotion screen bounding rectangle and 'wrapping around' to the other side of the bounding rectangle.

**Solution** : One way in which this problem can be solved is by removing the distortion effect from the promotion screen. This works; however, it is not the method I ended up going with.

Here is what I did within `NPSBDP_UIArmory_PromotionHero.ResizeScreenForBrigadierRank` :

1. I removed the promotion screen width increase since it really wasn't needed. The promotion screen was already well over 1700 pixels; furthermore, elongating a `UIScreen` seems to actually stretch the panels placed upon it.
2. I increased the width of a variety of background `UIPanel`'s, much like in the original promotion screen, to make way for the 8th rank column.
3. I shifted the whole promotion screen away from the left side of its bounding rectangle; this could be done since, with a shorter promotion screen, it would no longer sneak behind the soldier pawn.
4. I modified the location of the 8th rank column so it wouldn't sit on top of the 7th rank column.
5. I made a few other minor UI location adjustments.

Since the promotion screen's panels were no longer close to the bounding rectangle, the distortion effect could safely play without resulting in any panel tearing/wrapping.

**Final thoughts** :

Ultimately, I kept the original promotion screen, and made it the default, since many people will be used to its width and location.

Within `XComLW_UI.ini` I created a config variable, `USE_SHORTER_PROMOTION_SCREEN`, which, when enabled, substitutes the 'original' promotion screen with the 'shortened' promotion screen. This way users can choose to use the longer promotion screen with panel tearing or the shorter promotion screen without panel tearing.

**Problem** : Since the shorter promotion screen was moved to the right, to prevent it from hitting the bounding rectangle given to it by the 3D flash movie, Sparks and any other large pawn'd characters would overlap the 8th rank column.

**Solution** : The 'hero' promotion screen umap file, `3DUIBP_Armory_Promotion_Hero.umap`, was modified in the following ways :
1. The camera was rotated a bit, with respect to the Z-Axis, so that the soldier pawn appears farther to the right.
2. The screen object, within the umap, was moved along the Y axis so that it appears farther to the left.

**Problem** : The scroll bar would change position when switching from 'fullscreen' mode to 'windowed' mode; in other words, it would look ok in one mode but not the other.

**Solution** : It turns out that this problem only occurred when the scrollbar was being anchored to the screen; consequently, anchoring was removed and its X location was updated accordingly within `NPSBDP_UIArmory_PromotionHero.InitPromotion`. Now, the scrollbar remains on the right side of the promotion screen regardless of the screen mode.

Modifies : `NPSBDP_UIArmory_PromotionHeroColumn.uc`

The Flash rank label has now been replaced with 2 `UIText`s; this allows a label such as 'Master Sergeant' to appear on 2 lines. Additionally, to make room for this change, the rank icon and 'rank up' icon have been moved up.

**NOTE** : The rank label `UIText`s are filled according to this algorithm : 
1. The rank label is split into a string array, using a space as the delimiter.
2. The 1st word, if it exists, goes into the 1st `UIText`.
3. The 2nd word, if it exists, goes into the 2nd `UIText`.
_If there are more than 2 words, the 2nd `UIText` is a concatenation of the 2nd word to the 'nth' word separated by spaces._

### DIFF
--- a/LongWarOfTheChosen/Config/XComLW_UI.ini
+++ b/LongWarOfTheChosen/Config/XComLW_UI.ini
@@ -79,3 +79,6 @@ RECRUIT_FONT_SIZE_CTRL = 22					; Recruit screen font size for controller users.
 RECRUIT_Y_OFFSET_CTRL = -3					; Moves the stats up/down; negative numbers move up / positive numbers move down.
 RECRUIT_FONT_SIZE_MK = 20					; Recruit screen font size for mouse & keyboard users.
 RECRUIT_Y_OFFSET_MK = -2;					; Moves the stats up/down; negative numbers move up / positive numbers move down.
+
+[NewPromotionScreenByDefault_Integrated.NPSBDP_UIArmory_PromotionHero]
+USE_SHORTER_PROMOTION_SCREEN = false		; Use a slightly shorter promotion screen; this removes the panel tearing seen when cycling soldiers.
--- a/LongWarOfTheChosen/LongWarOfTheChosen.x2proj
+++ b/LongWarOfTheChosen/LongWarOfTheChosen.x2proj
@@ -21,6 +21,7 @@
     <Folder Include="Content\LW_Toolbox_Integrated\" />
     <Folder Include="Content\Maps\" />
     <Folder Include="Content\Missions\" />
+    <Folder Include="Content\NPSBD_Integrated" />
     <Folder Include="Content\Parcels\" />
     <Folder Include="Content\PCPs\" />
     <Folder Include="Content\Plots\" />
@@ -162,6 +163,12 @@
     <Content Include="Content\Missions\UMS_LWCommon.umap" />
     <Content Include="Content\Missions\UMS_LWContinuousReinforcements.umap" />
     <Content Include="Content\Missions\UMS_LWMissionTimer.umap" />
+    <Content Include="Content\NPSBD_Integrated\3DUIBP_Armory_Promotion_Hero.umap">
+      <SubType>Content</SubType>
+    </Content>
+    <Content Include="Content\NPSBD_Integrated\Readme.txt">
+      <SubType>Content</SubType>
+    </Content>
     <Content Include="Content\Plots\Plot_ALB_AlienBase_01_LW.umap" />
     <Content Include="Content\Plots\Plot_ALB_AlienBase_02_LW.umap" />
     <Content Include="Content\Plots\Plot_ALB_AlienBase_03_LW.umap" />
--- a/LongWarOfTheChosen/Src/NewPromotionScreenByDefault_Integrated/Classes/NPSBDP_UIArmory_PromotionHero.uc
+++ b/LongWarOfTheChosen/Src/NewPromotionScreenByDefault_Integrated/Classes/NPSBDP_UIArmory_PromotionHero.uc
@@ -25,12 +25,11 @@ var config array<int> FactionAbilityCosts;
 var config float BaseAbilityCostModifier;
 
 var config int ScrollbarPositionX;
+var config(LW_UI) bool USE_SHORTER_PROMOTION_SCREEN;
 
 // Position is the number by which we offset all ability indices.
 // 0 <= Position <= MaxPosition
 var int Position, MaxPosition;
-
-var int AdjustXOffset;
 
 simulated function ChangeSelectedColumn(int oldIndex, int newIndex)
 {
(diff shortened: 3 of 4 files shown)


pledbrook:
One issue with the shorter promotion screen is that SPARK pawns almost completely hide the Master Sergeant column at certain rotations.

kdm2k6:
When you have time try it out :

1. Make sure USE_SHORTER_PROMOTION_SCREEN within XComLW_UI is true, since it still relies on a config variable.
2. Place the unzipped folder into the 'Content' folder. [NPSBD_Integrated.zip](https://github.com/long-war-2/lwotc/files/5806296/NPSBD_Integrated.zip)

The umap seems to be the only way to make camera and screen changes within a 3D Flash movie.

**There are a couple small issues** :

1. When a soldier ranks up to Lance Corporal the 'New Rank' graphic overlaps the 'Select abilities' text a bit. I don't think this is a big deal since it only occurs at Lance Corporal and only until you choose an ability for the soldier.
2. When the spark uses certain guns, the bottom of the gun can get in the way of the 'Soldier AP' value. Although I like having the 'Soldier AP' value in the top right corner, I can move it, along with a few other UI elements, to the left a bit if necessary.

pledbrook (approved):
This looks good. I'll add the content file to git when I merge.

pledbrook:
Just tested with the normal promotion screen and the new NPSBD umap doesn't work well with it, as the left edge of the promotion panel is off screen. At this point, I'm questioning whether it's viable to have a flag for this. We probably need to commit either to the original or the shorter screen.

pledbrook:
@kdm2k6 There's an 'OverrideUIArmoryScale' highlander event that we could potentially use to shrink oversized pawns. Does that sound like a reasonable approach in place of the camera changes in the umap?

kdm2k6:
That could possibly work for the shorter promotion screen since, according to my comments, it sounds like the main reason for the camera changes were to prevent the pawns from blocking the promotion screen.

I'll take a look at some point; however, it might not be for a while as I'm working on other things and need to get myself back into a 'modding mood'.

pledbrook:
If you remember how these changes work, do you know if it's possible to shift the short promotion screen to the left while still fixing the tearing? I'm messing around with the X positions in `ResizeScreenForBrigadierRank()`, but can't get a half-way solution between the old promotion screen and the short one that works.

pledbrook:
Sadly, shrinking the pawns doesn't really fix the problem without a corresponding shift in the promotion screen. SPARKs are just too *wide*. To stop them from overlapping the MSGT column would require a scaling that makes them look like toys.

kdm2k6:
If I remember correctly, within `ResizeScreenForBrigadierRank`, `SetX()` moves the screen left or right; I moved it to the right via adding `ScreenXOffset`, as in `SetX(ScreenX + ScreenXOffset)`.

Individual elements are moved left or right via calls like : `MC.ChildSetNum("soldierAPLabel", "_x", MC.GetNum("soldierAPLabel._x") + RankColumnWidth);`.` soldierAPLabel` is the Flash name of an element on that screen. From my comments, it looks like the `bg` element needs to be moved a little bit differently from the other elements since it doesn't move with the rest of the screen when `SetX()` is called.

I believe that I moved the screen fairly close to the left edge while still preventing panel tearing; in other words, I'm not sure how much more the screen can be moved to the left without tearing beginning to appear.

pledbrook:
@kdm2k6 I've [moved this to the Community Promotion Screen](https://github.com/X2CommunityCore/X2CommunityPromotionScreen/pull/45), which I'm planning to switch to soon in place of the integrated NPSBD. The shorter promotion screen isn't universally popular (mostly because of what it does to the column headers). Wondering if "One way in which this problem can be solved is by removing the distortion effect from the promotion screen" might be the better option? Although to be honest, I prefer the shorter version.

kdm2k6:
No problem. Thanks for letting me know.

pledbrook:
(As mentioned previously, moved to Community Promotion Screen)

Issue #18, opened 2019-02-17
Squad select is auto filling empty squads


pledbrook:
If a new, empty squad is created for a mission (because no other squad is available), then squad select should not be allocating any soldiers to it or displaying them.

ansabalirai:
I think there is an option if you have MCM installed to disable that autofilling. It seems to be working for me

pledbrook:
robojumper's Squad Select fires a 'rjSquadSelect_AllowAutoFilling' event that can be used to override this on a per-mission basis. We probably need to override on every mission, or at least every mission with a squad assigned (which I think is every mission in LW2 because it creates a squad if necessary).

pledbrook:
So I've added the event listener that overrides this behaviour, but it means you can't change id via Mod Config Menu. It seems that we can either support the config option, but leave the default as autofill enabled, or implement the listener, in which case you can never enable autofill. Bleh.

pledbrook:
Fixed

PR #1212, opened 2021-01-05
Fix when 'New Rank' graphic is shown on promotion screen

kdm2k6:
**Introduction** :
The 'New Rank' graphic, positioned over top of a given column's rank name and symbol, should be displayed if both of these conditions are true :

1. The soldier's current rank is equivalent to the given column's associated rank.
2. No ability has been chosen from the column being looked at in condition 1.

In other words, the 'New Rank' graphic should be displayed if a soldier has ranked up and no ability has been chosen at this rank.

**Problem** :
Unfortunately, there is a problem with `NPSBDP_UIArmory_PromotionHero.UpdateAbilityIcons_Override` whose return value informs us about condition 2. This function iterates over the 'visible' abilities for a given column and determines if any have been chosen/bought; however, it ignores abilities which have been made 'invisible' due to scrolling.

_Here is a simple example_. Let's say we scroll down and are currently viewing the bottom 2 soldier ability rows, the XCom ability row, and the pistol ability row. Furthermore, let's assume that we have already purchased the ability, within this column, located in the top soldier ability row. In such a situation `UpdateAbilityIcons_Override` should return true, due to having an ability in this column; however, it will, in fact, return false since the invisible ability row is never checked.

**Solution** :
What we really want `UpdateAbilityIcons_Override` to do is, for any given column, determine if one of its 'soldier abilities' have been chosen/bought while, at the same time, ignoring 'XCom abilities' and 'pistol abilities'. In order to accomplish this we :

1. Call `GetAbilitiesPerRank` in order to determine the number of 'soldier abilities' per rank for the class being dealt with.
2. Determine if any abilities, between row 1 and row `GetAbilitiesPerRank` (inclusive), have been chosen/bought from the column being looked at; here, we disregard ability visibility.

### DIFF
--- a/LongWarOfTheChosen/Src/NewPromotionScreenByDefault_Integrated/Classes/NPSBDP_UIArmory_PromotionHero.uc
+++ b/LongWarOfTheChosen/Src/NewPromotionScreenByDefault_Integrated/Classes/NPSBDP_UIArmory_PromotionHero.uc
@@ -309,127 +309,157 @@ function bool HasBrigadierRank()
 
 function bool UpdateAbilityIcons_Override(out NPSBDP_UIArmory_PromotionHeroColumn Column)
 {
-	local X2AbilityTemplateManager AbilityTemplateManager;
-	local X2AbilityTemplate AbilityTemplate, NextAbilityTemplate;
-	local array<SoldierClassAbilityType> AbilityTree, NextRankTree;
-	local XComGameState_Unit Unit;
-	local UIPromotionButtonState ButtonState;
-	local int iAbility;
-	local bool bHasColumnAbility, bConnectToNextAbility;
+	local bool HasColumnAbility, ConnectToNextAbility, ColumnAboveUnitRank;
+	local int AbilityIndex, i, NumberOfSoldierAbilities;
 	local string AbilityName, AbilityIcon, BGColor, FGColor;
-//	local int NewMaxPosition;
-
+	local array<SoldierClassAbilityType> AbilityTree, NextRankTree;
+	local UIPromotionButtonState ButtonState;
+	local X2AbilityTemplate AbilityTemplate, NextAbilityTemplate;
+	local XComGameState_Unit Unit;
+	local X2AbilityTemplateManager AbilityTemplateManager;
+	
 	AbilityTemplateManager = class'X2AbilityTemplateManager'.static.GetAbilityTemplateManager();
 	Unit = GetUnit();
 	AbilityTree = Unit.GetRankAbilities(Column.Rank);
+	ColumnAboveUnitRank = (!RevealAllAbilities && Column.Rank >= Unit.GetRank());
 
-//	NewMaxPosition = Max(AbilityTree.Length - NUM_ABILITIES_PER_COLUMN, NUM_ABILITIES_PER_COLUMN);
-//	if (NewMaxPosition > MaxPosition)
-//		MaxPosition = NewMaxPosition;
-
-	// MaxPosition is the maximum value for Position
+	// MaxPosition is the maximum value for Position.
 	MaxPosition = Max(AbilityTree.Length - NUM_ABILITIES_PER_COLUMN, MaxPosition);
 
-	//`LOG("MaxPosition" @ MaxPosition,, 'PromotionScreen');
 	Column.AbilityNames.Length = 0;
 	
-	//for (iAbility = 0; iAbility < AbilityTree.Length; iAbility++)
-	//{
-	//	`LOG("Create Column" @ Column.Rank @ AbilityTree[iAbility].AbilityName,, 'PromotionScreen');
-	//}
-
-
-	for (iAbility = Position; iAbility < Position + NUM_ABILITIES_PER_COLUMN; iAbility++)
+	for (AbilityIndex = Position; AbilityIndex < Position + NUM_ABILITIES_PER_COLUMN; AbilityIndex++)
 	{
-		AbilityTemplate = AbilityTemplateManager.FindAbilityTemplate(AbilityTree[iAbility].AbilityName);
+		AbilityTemplate = AbilityTemplateManager.FindAbilityTemplate(AbilityTree[AbilityIndex].AbilityName);
 		
 		if (AbilityTemplate != none)
 		{
 			if (Column.AbilityNames.Find(AbilityTemplate.DataName) == INDEX_NONE)
 			{
 				Column.AbilityNames.AddItem(AbilityTemplate.DataName);
-				//`LOG(iAbility @ "Column.AbilityNames Add" @ AbilityTemplate.DataName @ Column.AbilityNames.Length,, 'PromotionScreen');
 			}
 
-			// The unit is not yet at the rank needed for this column
-			if (!RevealAllAbilities && Column.Rank >= Unit.GetRank())
+			// The unit is not yet at the rank needed for this column.
+			if (ColumnAboveUnitRank)
 			{
 				AbilityName = class'UIUtilities_Text'.static.GetColoredText(m_strAbilityLockedTitle, eUIState_Disabled);
 				AbilityIcon = class'UIUtilities_Image'.const.UnknownAbilityIcon;
 				ButtonState = eUIPromotionState_Locked;
 				FGColor = class'UIUtilities_Colors'.const.BLACK_HTML_COLOR;
 				BGColor = class'UIUtilities_Colors'.const.DISABLED_HTML_COLOR;
-				bConnectToNextAbility = false; // Do not display prereqs for abilities which aren't available yet
+				// Do not display prerequisites for abilities which aren't available yet.
+				ConnectToNextAbility = false;
 			}
-			else // The ability could be purchased
+			// The ability can, potentially, be purchased.
+			else
 			{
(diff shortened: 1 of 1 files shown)


LongWarOfTheChosen/Src/NewPromotionScreenByDefault_Integrated/Classes/NPSBDP_UIArmory_PromotionHero.uc
```diff
@@ -309,127 +309,146 @@ function bool HasBrigadierRank()
...
+	// be false since the invisible ability row is never checked.
+	//
+	// What we really want to do is, for this particular column, determine if one of its 'soldier abilities' 
+	// have been chosen/bought while, at the same time, ignoring 'XCom abilities' and 'pistol abilities'. 
+	// Since 'soldier abilities' are located in the top 3 rows, we simply need to see if any of them have been 
+	// chosen/bought regardless of visibility.
+	HasColumnAbility = false;
+	if (!ColumnAboveUnitRank)
+	{
+		for (i = 0; i < 3; i++)
```
pledbrook, line 440: What's the 3 for here? If this is the number of class abilities available at each rank, then the value needs to be pulled from `GetAbilitiesPerRank(UnitState)`.

pledbrook (changes requested):

LongWarOfTheChosen/Src/NewPromotionScreenByDefault_Integrated/Classes/NPSBDP_UIArmory_PromotionHero.uc
kdm2k6: I believe the 'New Rank' graphic should only appear if someone has not chosen a 'soldier ability' at their given rank. It should ignore whether or not an 'XCom' or 'Pistol' ability has been chosen at this rank.

- The 1st, 2nd, and 3rd abilities for a given rank correspond to the 1st, 2nd, and 3rd rows in rank column; these are the 'soldier abilities'. 
- The 4th ability for a given rank corresponds to the 4th row in a rank column, which is the 'XCom' ability. 
- The 5th ability, if it exists, for a given rank corresponds to the 5th row in a rank column, which is the 'Pistol' ability.

This checks to see if any 'soldier abilities' have been selected at their given rank.

Another way to do it would be to cycle through all class abilities, at a given rank, and only check 'soldier abilities'. Is there a way to determine if an ability is a 'soldier ability' vs 'XCom ability' vs 'Pistol ability'. This way ordering would be unimportant.

pledbrook: See `GetAbilitiesPerRank()`. It's important not to use 3 because some classes may have only 2 or maybe more than 4. Depends on what soldier classes are added by mods.

kdm2k6: Right. That should give the total number of abilities for each rank; however, is there any way to find out "the type" of ability being looked at? 

Let's say I'm looking at a Sharpshooter at Lance Corporal rank. I can see 3 soldier abilities : _Death from Above_, _Rapid Targeting_, and _Snap Shot_. Furthermore, I see 1 XCom ability, _Flashbanger_, and 1 pistol ability, _Return Fire_.

For this modification to work properly, I would need to be able to determine if an ability is a base soldier ability or "some other" ability.

I believe in the base game the 'New Rank' graphic would show whenever an ability had not been chosen for the soldier's rank. However, things are more complicated in LW. Looking at the example above, let's say the pistol ability _Return Fire_ was bought; I still think 'New Rank' should show since a base soldier ability has not been bought. XCom abilities and pistol abilities seem like 'extras' but not 'core' upgrades.

Worst case, if this is not possible, I can just loop over "all" abilities like mentioned previously, and see if any have been bought.

pledbrook: Yeah, sorry, when I glanced at it earlier I thought it was giving just the class abilities. I think it might be best to rely on the promotion UI config that says how many class abilities there are. `ClassAbilitiesPerRank`. If soldier class mods want to support the new promotion UI/LWOTC, they will need to add entries to that config array. I think I'm using it elsewhere already.

kdm2k6: Can we assume that class abilities always come before XCom/Pistol/Non-Core abilities? I'm guessing that class abilities will always take up the first 'n' rows while 'other' abilities will begin at row 'n+1'.

pledbrook: That's the assumption I've worked on before. It's not foolproof, because I'm pretty sure it depends on the order in which they're declared in XComClassData.ini. Also, this probably wouldn't work at all for RPGO, but that's fine. I think just having a rule that LWOTC soldier classes have their class abilities first should work out OK.

kdm2k6: Within XComClassData.ini, I think a lot of AbilitySlots have an extra 'none' ability at the end.

For example : 
```
+SoldierRanks=(AbilitySlots=((AbilityType=(AbilityName="RevivalProtocol", ApplyToWeaponSlot=eInvSlot_SecondaryWeapon)), \\
(AbilityType=(AbilityName="Sentinel_LW", ApplyToWeaponSlot=eInvSlot_PrimaryWeapon)), \\
(AbilityType=(AbilityName="CombatProtocol", ApplyToWeaponSlot=eInvSlot_SecondaryWeapon)), \\
(RandomDeckName="Tier1_XComAbilities"), \\
(AbilityType=(AbilityName="ReturnFire", ApplyToWeaponSlot=eInvSlot_Pistol)), \\
***_THE_COMMA_JUST_BEFORE_THIS_STATEMENT_CREATES_AN_EXTRA_ELEMENT***), \\
aStatProgression=((StatType=eStat_Offense,StatAmount=2), (StatType=eStat_Will,StatAmount=5), (StatType=eStat_Hacking,StatAmount=10), (StatType=eStat_HP,StatAmount=1), (StatType=eStat_Strength,StatAmount=0), (StatType=eStat_CombatSims,StatAmount=0))\\
)
```

I noticed that when you log `AbilityTree.Length` after setting it via :
`AbilityTree = Unit.GetRankAbilities(Column.Rank);`
you often get 6 even though there are 5 abilities for a given rank.

I'm not sure if this matters or not. I'm assuming these 'none' abilities are just skipped in code generally speaking.

I will post it in issues and it can be closed if it is a non-issue.

LongWarOfTheChosen/Src/NewPromotionScreenByDefault_Integrated/Classes/NPSBDP_UIArmory_PromotionHero.uc
```diff
@@ -309,127 +309,157 @@ function bool HasBrigadierRank()
...
+		// associated AbilitiesPerRank value is returned. AbilitiesPerRank represents the number of 'soldier abilities'
+		// per rank for a particular soldier class.
+		// 2.] If ClassAbilitiesPerRank has NOT been defined for the unit's class then the function determines
+		// the number of abilities for each rank with respect to that class. The largest 'abilities for a given
+		// rank' value is then returned.
+		NumberOfSoldierAbilities = GetAbilitiesPerRank(Unit);
+		if (NumberOfSoldierAbilities > AbilityTree.Length)
+		{
+			NumberOfSoldierAbilities = AbilityTree.Length;
+		}
```
pledbrook, line 449:
```suggestion
		NumberOfSoldierAbilities = Min(GetAbilitiesPerRank(Unit), AbilityTree.Length);
```

LongWarOfTheChosen/Src/NewPromotionScreenByDefault_Integrated/Classes/NPSBDP_UIArmory_PromotionHero.uc
```diff
@@ -309,127 +309,157 @@ function bool HasBrigadierRank()
...
+
+		for (i = 0; i < NumberOfSoldierAbilities; i++)
+		{
+			AbilityTemplate = AbilityTemplateManager.FindAbilityTemplate(AbilityTree[i].AbilityName);
+			if (AbilityTemplate != none && Unit.HasSoldierAbility(AbilityTemplate.DataName))
```
pledbrook, line 454: Why is the template needed? Would `Unit.HasSoldierAbility(AbilityTree[i].AbilityName)` not work?

pledbrook (changes requested):
Provided a suggestion and I have one question.

pledbrook:
This was fixed a different way in the Community Promotion Screen for its [issue #18](https://github.com/X2CommunityCore/X2CommunityPromotionScreen/issues/18).

PR #1213, opened 2021-01-05
Fix edge case generating invalid soldier

PointMeAtTheDawn:
It's pretty unlikely to happen, but this prevents a 0 hp soldier from being generated.

Previously if iterations ever reached 1000, all remaining stat swaps would be immediately processed regardless of whether they were valid. Now we will stop processing stat swaps when we reach our limit.

This fix could fractionally lower our chance of performing a ComInt swap, so I also moved that to be processed first so that COMINT_BASE_SWAP_CHANCE is more authoritative.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Toolbox_Integrated/Classes/XComGameState_Unit_LWRandomizedStats.uc
+++ b/LongWarOfTheChosen/Src/LW_Toolbox_Integrated/Classes/XComGameState_Unit_LWRandomizedStats.uc
@@ -99,20 +99,6 @@ function RandomizeInitialStats(XComGameState_Unit Unit)
 		ComIntTotalWeight += ComIntSwap.Weight;
 	}
 
-	//randomly apply a bunch of stat swaps to get starting stat offset
-	for(idx = 0; idx < NumSwaps; idx++)
-	{
-		do {
-			Swap = SelectRandomStatSwap(TotalWeight);
-		} until (IsValidSwap(Swap, Unit) || (++iterations > 1000));
-
-		if (iterations > 1000)
-		{
-			break;
-		}
-		CharacterInitialStats_Deltas[Swap.StatUp] += Swap.StatUp_Amount;
-		CharacterInitialStats_Deltas[Swap.StatDown] -= Swap.StatDown_Amount;
-	}
 	//Roll a chance for a stat swap and don't roll combat intelligence for rebels
 	if (`SYNC_RAND(100) < default.COMINT_BASE_SWAP_CHANCE && InStr(Unit.GetMyTemplateName(), "Rebel") != 0 )
 	{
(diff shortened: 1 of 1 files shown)


PointMeAtTheDawn:
[Force push squashes the third commit for a cleaner commit log is all, unfortunate that the fork PR doesn't reflect this.]

LongWarOfTheChosen/Src/LW_Toolbox_Integrated/Classes/XComGameState_Unit_LWRandomizedStats.uc
```diff
@@ -99,26 +99,32 @@ function RandomizeInitialStats(XComGameState_Unit Unit)
 		ComIntTotalWeight += ComIntSwap.Weight;
 	}
 
+	//Roll a chance for a stat swap and don't roll combat intelligence for rebels
+	if (`SYNC_RAND(100) < default.COMINT_BASE_SWAP_CHANCE && InStr(Unit.GetMyTemplateName(), "Rebel") != 0 )
+	{
+		do 
+		{
+			ComIntSwap = SelectComIntStatSwap(ComIntTotalWeight);
+		} until (IsValidComIntSwap(ComIntSwap, Unit) || (++iterations > 1000));
```
pledbrook, line 108: If we're going to put this first, we should reduce the number of potential iterations, otherwise a failure here will block any other stat swaps. Alternatively, reset `iterations` between the loops.

LongWarOfTheChosen/Src/LW_Toolbox_Integrated/Classes/XComGameState_Unit_LWRandomizedStats.uc
```diff
@@ -99,26 +99,32 @@ function RandomizeInitialStats(XComGameState_Unit Unit)
...
+		if (iterations <= 1000)
+		{
+			CharacterInitialStats_Deltas[ComIntSwap.StatSwapped] += ComIntSwap.StatSwapped_Amount;
+			CharacterComIntDelta = ComIntSwap.ComIntStatchange;	
+		}
+	}
```
pledbrook, line 114:
```suggestion
	}

```

LongWarOfTheChosen/Src/LW_Toolbox_Integrated/Classes/XComGameState_Unit_LWRandomizedStats.uc
```diff
@@ -99,26 +99,32 @@ function RandomizeInitialStats(XComGameState_Unit Unit)
...
 	//randomly apply a bunch of stat swaps to get starting stat offset
 	for(idx = 0; idx < NumSwaps; idx++)
 	{
 		do {
 			Swap = SelectRandomStatSwap(TotalWeight);
 		} until (IsValidSwap(Swap, Unit) || (++iterations > 1000));
-
+		if (iterations > 1000)
```
pledbrook, line 121: Considering the number of times `1000` is used, I think it's about time we assigned it to a constant, say `MAX_SWAP_ATTEMPTS`.

pledbrook (changes requested):
Thanks for this. See comments/suggestions inline.

pledbrook:
This pull request is out of date with the current code, so I applied the basic fix from it in [this commit](https://github.com/long-war-2/lwotc/commit/25e2caa51565699900b0ba92b6379b16c1178895).

PR #1214, opened 2021-01-06
Option to restrict more classes from officer training

K1ruka:
Added feature for better compatibility with mods like Playable Aliens and MEC Troopers, whose units shouldn't really be Officers due partly to missing animations.


K1ruka:
Restored branch.

pledbrook (approved):
Thanks for this!

PR #1219, opened 2021-01-12
Update translation fra v22 3

Savancosinus:
Localization's update.

FRA:
- Translation of the whole mod.
- I corrected some mistakes that were here since... The release of the original mod (LW2). :) (Examples: Laser sight, render of elerium core, etc.).
- I cleaned up some descriptions. Example: Cutthroat -> The french version describes a "%" crit bonus, but it's a "flat" buff. I think that's could be a little bit confusing with "ratio" bonus (like Deadeye). This issue doesn't appear in the english version.

INT:
- Update of "[BasicStockProject X2TechTemplate]" +(ADV/SUP version). I suppressed "Any troop that equips a stock gains a chance to convert misses to grazes", because that's not the case anymore.
- I corrected some typos (missing quotes, etc.).
- Add "LWotC Needs Translation" on "[Bombard_LW X2AbilityTemplate]", because you suppressed the word "claymore".

EDIT: I'll push an add-on in the next few days, to translate the changes that were committed the last days. I'll also add a cool name for "Lead the target", as I can't find any satisfying name for the moment. :/


Savancosinus:
Ok, FRA translation is up to date with the version 23.
Hope you could merge it before you release the beta 4!

PR #1222, opened 2021-01-14
Change priority of Flashbanger and Smoker templates

GafrieGithub:
Currently the interaction between Smoker/Flashbanger and Full Kit abilities depend on the order of which they were acquired. So, Full Kit is inconsistent in granting or not extra flashbang/smoke grenade. Additionally, per Full Kit description it is arguable that it should not apply to these extra grenades at all.
This changes priorities of Smoker/Flashbanger templates, so that they happen after Full Kit and therefore would not have extra charges.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_PerkPack_Integrated/Classes/X2Ability_PerkPackAbilitySet2.uc
+++ b/LongWarOfTheChosen/Src/LW_PerkPack_Integrated/Classes/X2Ability_PerkPackAbilitySet2.uc
@@ -367,8 +367,9 @@ static function TrojanVirusVisualizationRemoved(XComGameState VisualizeGameState
 //this ability grants a free equip of a flashbang grenade
 static function X2AbilityTemplate AddFlashbanger()
 {
-	local X2AbilityTemplate			Template;
-	local X2Effect_TemporaryItem	TemporaryItemEffect;
+	local X2AbilityTemplate						Template;
+	local X2Effect_TemporaryItem				TemporaryItemEffect;
+	local X2AbilityTrigger_UnitPostBeginPlay	Trigger;
 
 	`CREATE_X2ABILITY_TEMPLATE(Template, 'Flashbanger');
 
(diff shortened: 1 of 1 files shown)


LongWarOfTheChosen/Src/LW_PerkPack_Integrated/Classes/X2Ability_PerkPackAbilitySet2.uc
```diff
@@ -412,7 +418,11 @@ static function X2AbilityTemplate AddSmokeGrenade()
 	Template.Hostility = eHostility_Neutral;
 	Template.AbilityToHitCalc = default.DeadEye;
 	Template.AbilityTargetStyle = default.SelfTarget;
-	Template.AbilityTriggers.AddItem(default.UnitPostBeginPlayTrigger);
+
+	Trigger = new class'X2AbilityTrigger_UnitPostBeginPlay';
+	Trigger.Priority -= 20; // delayed so that Full Kit happen first
```
Grobobobo, line 423: Hold on, I'm pretty sure lower priority = goes faster, you sure you tested it?

Grobobobo (changes requested):

LongWarOfTheChosen/Src/LW_PerkPack_Integrated/Classes/X2Ability_PerkPackAbilitySet2.uc
GafrieGithub: I believe that I did test it at the time. 
For example here is the line form Dense Smoke ability:
`Trigger.Priority -= 40; // delayed so that any other abilities that add items happen first`
Or this line from vanilla HotLoadAmmo():
`PostBeginPlayTrigger.Priority -= 10;        // Lower priority to guarantee ammo modifying effects (e.g. Deep Pockets) already run.`
So I believe higher priority should happen earlier, as the name suggests.

Grobobobo:
It confused me because if you look at chosen turn priority -100 priority makes them go faster. Maybe trigger priority works differently though.

PR #1223, opened 2021-01-15
Improving descriptions further

GafrieGithub:
These are further changes to abilities descriptions.
Generally, changes were focused on using variable config values in the descriptions instead of plain text values and making sure that the abilities fully and correctly describe what they do. There are couple of cases where ability descriptions were not fully matched to their implementations, since I believe their current behaviours were not intended and should be fixed instead. For these, you may further refer to Sting Grenades/Concussion Rocket/Full Kit commit comments.
There are also some typo and wording issues fixed.

Finally, I have to admit that since English is not my native language and there were cases where I had to add some lines from scratch, there may be some grammar issues or somesuch. So better keep an eye out for those.


Grobobobo:
I've got permission to merge things directly to master, so switch the PR to that branch, and resolve the merge conflicts.

PR #1224, opened 2021-01-15
Remove "Mental" damage type from Sting Grenades and Concussion Rocket

GafrieGithub:
Currently, the Sting Grenades replacement items are created with their disorient and stun effects set to have "Mental" damage type (which is the default value of CreateDisorientedStatusEffect() and CreateStunnedStatusEffect() functions). This is unlike normal flashbang grenades, which are created without this damage type. Therefore, it leads to the situation where Grenadier who bought Sting Grenades perk suddenly can't disorient certain enemies which he previously could, for example Spectres.
This change make their sting grenades grenades consistent with flashbangs, if a bit stronger (which they probably did not need, but oh well).

As for the Concussion Rocket, it had only disorient effect created with default arguments, so you could stun but not disorient Spectres, for no apparent reason. This change make it consistent as well.

Additionally, I want to mention the explicit immunities for PsiZombie, AdvPsiWitchM2 and AdvPsiWitchM3 to all effects of Concussion Rocket, including damage. I found it absolutely counter-intuitive (damage immunity in particular), and don't know why it have to be there, but I left it there for others to decide whether it should be fixed design-wise.


Grobobobo:
If you can, please change the branch again.

PR #1245, opened 2021-02-06
Move some abilities assignments from characters to weapons

GafrieGithub:
Changed "LockedOn", "CloseCombatSpecialist", "CloseAndPersonal", "BringEmOn", "Executioner_LW" and  "CoupdeGrace2" from being assigned character templates to being assigned to weapon templates.

For "CloseCombatSpecialist", "CoupdeGrace2" and "BringEmOn" it is required for Ability:BOUND_WEAPON_NAME tag in the descriptions to work correctly. Additionally, as far as I can tell BringEmOn didn't even work properly the way it was set up.

For "LockedOn" and "Executioner_LW" reassignment doesn't strictly matter, but given that these abilities were assigned to weapons for other units (Gunners and Najas respectively) already, it seemed proper to move these as well.

"CloseAndPersonal" being assigned to weapon does not seem to change anything and was changed only because similar "// weapon?" comment which implied it was the intent at some point. Plus it makes more sense this way.


Grobobobo:
Peter wants to avoid assigning abilities to weapon template because it makes it messy-ish. The assignment itself is done through PrimaryWeaponAbilities and SecondaryWeaponAbilities config in LW_Overhaul.ini.

PR #1261, opened 2021-03-03
More Psi abilities for Faceless chance reduction

K1ruka:
The edited line now checks for one of the following abilities: Mind Merge, Soulfire, or Insanity for Faceless chance reduction, rather than just Mind Merge. This is for better compatibility with alternate Psi setups such as Psionics Overhaul V3 with Mod Jam.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/XComGameState_LWOutpost.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/XComGameState_LWOutpost.uc
@@ -269,10 +269,14 @@ function StateObjectReference CreateRebel(XComGameState NewGameState, XComGameSt
 			{
 				FacelessChance *= 0.6;
 			}
-			if (Unit.HasAbilityFromAnySource.Find(PsionAbilityName))
-			{
-				FacelessChance *= 0.6;
-			}
+       			foreach default.FacelessReductionPsiAbilities(PsionAbilityName)
+            		{
+                		if (Unit.HasAbilityFromAnySource(PsionAbilityName))
+                		{
+                    			FacelessChance *= 0.6;
+					break;
+                		}
+            		}
 			if (Unit.HasItemOfTemplateType('Battlescanner'))
 			{
 				FacelessChance *= 0.6;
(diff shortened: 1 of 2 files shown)


LongWarOfTheChosen/Src/LW_Overhaul/Classes/XComGameState_LWOutpost.uc
```diff
@@ -265,7 +265,7 @@ function StateObjectReference CreateRebel(XComGameState NewGameState, XComGameSt
 			{
 				FacelessChance *= 0.6;
 			}
-			if (Unit.HasSoldierAbility('MindMerge'))
+			if (Unit.HasSoldierAbility('MindMerge') || Unit.HasSoldierAbility('Soulfire') || Unit.HasSoldierAbility('Insanity'))
```
Grobobobo, line 268: If you're doing this, wouldn't be best to make it configurable at this point? I think there can be a batter implementation there.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/XComGameState_LWOutpost.uc
```diff
@@ -265,9 +269,12 @@ function StateObjectReference CreateRebel(XComGameState NewGameState, XComGameSt
 			{
 				FacelessChance *= 0.6;
 			}
-			if (Unit.HasSoldierAbility('MindMerge'))
+			foreach default.FacelessReductionPsiAbilities(PsionAbilityName)
```
Grobobobo, line 272: This time the multiplier will trigger for EACH of those abilities. Use find instead of the loop.

Grobobobo (changes requested):

LongWarOfTheChosen/Src/LW_Overhaul/Classes/XComGameState_LWOutpost.uc
```diff
@@ -245,6 +248,7 @@ function StateObjectReference CreateRebel(XComGameState NewGameState, XComGameSt
 	local array<X2StrategyElementTemplate> PersonalityTemplates;
 	local StateObjectReference NewUnitRef;
 	local float FacelessChance;
+	local name PsionAbilityName;
```
Grobobobo, line 251: Where is that set? What's the Connection between PsionAbilityName and FacelessReductionPsiAbilities?

Grobobobo (changes requested):
Pretty sure this still doesn't work.

pledbrook (commented):
I'm not a huge fan of the arbitrary cap on the number of abilities that can have an effect. I think I'll have a go at implementing a different algorithm that makes additional abilities contribute to the reduction, but not by much. So two abilities should be noticeably better than 1, but beyond that, it's not really worth adding more.

pledbrook:
Added in [this commit](https://github.com/long-war-2/lwotc/commit/54ff13e4bbe999fb6ef5dd4acae0ae5a12116cf5).

PR #1265, opened 2021-03-14
Added aliens to Rendezvous spawns

Schizosaccaromyces:
1. Split the encounter lists for Leader and Follower (previously both drew from the same Encounter List)
2. Spectres, Archons and Sectoids can now spawn as Leaders and Followers
3. Vipers and Faceless can spawn as Followers
4. Muton Elites, Centurions, and Sectoid Commanders can spawn as Leaders
5. Officers and Priests now only spawn as Leaders
6. Trooper variants and Stun Lancers now only spawn as Followers (Sergeant included)

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/CHItemSlot_PistolSlot_LW.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/CHItemSlot_PistolSlot_LW.uc
@@ -60,9 +60,9 @@ static function bool CanAddItemToPistolSlot(
     WeaponTemplate = X2WeaponTemplate(Template);
     if (WeaponTemplate != none)
     {
-        return default.PISTOL_SLOT_WEAPON_CLASSES.Find(WeaponTemplate.WeaponCat()) == INDEX_NONE;
+        return WeaponTemplate.WeaponCat == 'pistol';
     }
-    return true;
+    return false;
 }
 
 static function bool HasPistolSlot(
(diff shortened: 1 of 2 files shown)


Grobobobo:
What about andromedons? don't feel like it?

Schizosaccaromyces:
More that they are not exactly what I would bring if I wanted to be quiet under the Resistance. Muton Centurions and Elites make sense up to a point (they are smarter than the rank and file) but Andromedons are huge clanky things. 
Same reason I did not put zerkers (though codices is also for Golden Path progression issues)

Schizosaccaromyces:
I´ve added by mistake two commits to CHItemSlot_PistolSlot_LW and XComLW_Overhaul.ini to allow other item categories to be added into the pistol slot. Those are unrelated to the subject of the PR so they can be reviewed/discarded separately. Apologies for the mess.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/CHItemSlot_PistolSlot_LW.uc
```diff
@@ -59,9 +60,9 @@ static function bool CanAddItemToPistolSlot(
     WeaponTemplate = X2WeaponTemplate(Template);
     if (WeaponTemplate != none)
     {
-        return WeaponTemplate.WeaponCat == 'pistol';
+        return default.PISTOL_SLOT_WEAPON_CLASSES.Find(WeaponTemplate.WeaponCat()) == INDEX_NONE;
```
Grobobobo, line 63: Shouldn't it be != INDEX_NONE?

Grobobobo (changes requested):
Ideally You'd revert those commits on this branch and submit the PR somewhere else.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/CHItemSlot_PistolSlot_LW.uc
Schizosaccaromyces: Could be? Would != INDEX_NONE mean that "if it´s not on the index, do this"? I thought it would be the other way around but now I am not so sure. 
Anyhow, will revert these and open a new branch (not familiar with Github at all so I thought it would do it automatically on a new PR)

Grobobobo: INDEX_NONE means that the thing has not been found in an array.

PR #1268, opened 2021-03-16
Apply Insider Knowledge continent bonus to all weapon mods

GafrieGithub:
Currently Insider Knowledge continent bonus only applies to Laser Sights, Auto-Loaders and Expanded Magazines (because they are functionally unchanged from vanilla). This changes the other weapon mods (Scopes,, Hair Triggers, Stocks and Suppressors) so that Insider Knowledge improves them too, like it says it should.

It may be argued that these weapon mods or this continent bonus do not need buffs, so the bonus values I chose are small and might be overall insignificant.  But the principle of continent bonus actually working is still there and at least bonus values could be changed later on, if one would want to.


Grobobobo:
I think we're more likely to remove this bonus, but I'm not sure.

GafrieGithub:
> 
> 
> I think we're more likely to remove this bonus, but I'm not sure.

Sure, but I believe that is the balance consideration and it may or may not happen. Right now though the bonus is there and is not doing what it says it does. The option to disable the bonus later will still be there.

pledbrook:
Thanks!

PR #1272, opened 2021-03-20
Prevent Field Surgeon to apply to bleeding out soldiers

GafrieGithub:
Current description of Field Surgeon says: "It does not affect soldiers who suffered killing wounds and are stabilized from bleeding out.". However, it is erroneus for several reasons:
1. The code in X2Effect_FieldSurgeon checks only for BleedingOut status and not for Unconscious for target soldier.
2. BleedingOut status would be explicitly cleared prior to the Field Surgeon checks, if the soldier was carried out by someone else during evacuation (which is often the case).
3. Most importantly, both BleedingOut and Unconscious effects are cleared when the unit leaves play, which is way before Field Surgeon checks happen.

This changes X2Effect_FieldSurgeon effect to store the fact that soldier was bleeding out during any point of the mission and use this knowledge later to check whether Field Surgeon should apply.

To note. there are checks in the code for whether the specialist himself is unconscious/bleeding out. It supposed to prevent Field Surgeon to apply in these cases. However, for the same reasons these checks do nothing. This is not fixed in this commit. It might be okay, since Field Surgeon description does not mention this requirement in the first place, however it can lead to somewhat strange behavior where if the specialist himself was bleeding out he will heal everyone else, but not himself.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_PerkPack_Integrated/Classes/X2Effect_FieldSurgeon.uc
+++ b/LongWarOfTheChosen/Src/LW_PerkPack_Integrated/Classes/X2Effect_FieldSurgeon.uc
@@ -10,6 +10,7 @@ class X2Effect_FieldSurgeon extends X2Effect_Persistent config(LW_SoldierSkills)
 
 var config array<int> FIELD_SURGEON_CHANCE_FOR_NUM_EFFECTS;
 var name FieldSurgeonAppliedUnitValue;
+var name FieldSurgeonUnitWasBleedingOut;
 
 function RegisterForEvents(XComGameState_Effect EffectGameState)
 {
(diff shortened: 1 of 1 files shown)


Grobobobo:
Wouldn't it be better to just make it officially work on bleedouts to simplify the ability?

GafrieGithub:
> 
> 
> Wouldn't it be better to just make it officially work on bleedouts to simplify the ability?

Probably. From the viewpoint of keeping the original intent it is surely a bug, but if you want to keep it simpler instead then removing that one line from the description is surely the way to go. I chose to fix the implementation rather than description, to try keeping the intent, but ultimately I think it is not for me to choose the general approach. 
Balance-wise it probably does not change that much either way, but I believe Field Surgeon is already the most common pick at that rank, so maybe it not working on bleedouts tone it down slightly?

PR #1273, opened 2021-03-25
Make Field Surgeon apply to liaisons

GafrieGithub:
Currently Field Surgeon effect does not get applied to liaison (or rebels for that matter) on retaliation missions. This happen because Field Surgeon relies on UnitPostBeginPlayTrigger to apply its effect on all allies. And the liaison and rebels get added onto the map after that. (The relevant order of things are in X2TacticalGameRuleset.BeginTacticalPlay() function: first it applies UnitPostBeginPlayTrigger for all abilities, and then triggers OnTacticalBeginPlay event which LW2 code subscribe to to get liaison added).

This changes Field Surgeon effect to use OnUnitBeginPlay event to apply X2Effect_FieldSurgeon. on all relevant units. I have to admit that it took me an embarassing amount of time to get it working as it should. And even now I am not fully sure I've done it the proper way (though I tested that it works).

Another things of note. There are other abilities that currently have the same problem: Deflade, Combined Arms, Fire Discipline and Esprit de Corps. Curiously, unlike other officer abilities, Lead by Example seems to trigger on UnitMoveFinished event in addition to UnitPostBeginPlay and therefore probably avoid this issue.  Maybe I should have used its approach instead of  using OnUnitBeginPlay, I don't know what is the best way.  I don't plan to similarly fix aforementioned officer auras until after the change I made to Field Surgeon would be deemed problem-free.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_PerkPack_Integrated/Classes/X2Ability_PerkPackAbilitySet2.uc
+++ b/LongWarOfTheChosen/Src/LW_PerkPack_Integrated/Classes/X2Ability_PerkPackAbilitySet2.uc
@@ -919,6 +919,8 @@ static function X2AbilityTemplate AddFieldSurgeon()
 {
 	local X2AbilityTemplate						Template;
 	local X2Effect_FieldSurgeon					FieldSurgeonEffect;
+	local X2AbilityTrigger_EventListener 		EventListener;	
+	local X2Condition_UnitProperty              TargetProperty;
 
 	`CREATE_X2ABILITY_TEMPLATE(Template, 'FieldSurgeon');
 	Template.IconImage = "img:///UILibrary_LW_PerkPack.LW_AbilityFieldSurgeon";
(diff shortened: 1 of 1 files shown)


Grobobobo (changes requested):
Sorry, but this code does not make sense to me, by removing the multi target you're making field surgeon apply to only the unit that has the ability.

GafrieGithub:
> 
> 
> Sorry, but this code does not make sense to me, by removing the multi target you're making field surgeon apply to only the unit that has the ability.

OnUnitBeginPlay will trigger for every unit on the battlefield, which naturally includes every soldier and liaison. Therefore AbilityTriggerAgainstSingleTarget in FieldSurgeonOnUnitBeginPlay will trigger for every such unit applying the FielsSurgeon effect, provided that the unit satisfy AbilityTargetConditions defined earlier ro exclude enemies etc.

Grobobobo:
Yes, on unitbeginplay event happens for every unit, but the ability trigger shouldvonly work work for the unit that has the ability.

GafrieGithub:
I believe that 
`Template.AbilityTargetStyle = default.SingleTargetWithSelf`
make it that it can target any unit with the effect and 
`SourceAbilityState.AbilityTriggerAgainstSingleTarget(UnitState.GetReference(), false);`
make it trigger against the unit which triggered OnUnitBeginPlay.

If the question is whether I tested it applying to other units, then, yes, I did. It did indeed applied the effect to all soldiers/specialist himself/liaison.

PR #1280, opened 2021-03-30
Allow items specified as Pistol Slot to show in Armory UI for soldiers (part 2)

Schizosaccaromyces:
I did not edit the check for allowing the item to show on the Inventory UI (so whilst soldiers could equip the specified items, they would not show in the Armory UI). I confirmed this by having a Templar with an Autopistol equipped on the secondary despite those being in eINVSlot_Pistol (Autopistol Overhaul allows sidearms and pistols to go through both slots.


Schizosaccaromyces:
Alright, I am not sure what I did but this doesn´t work if I load the script together with LWOTC. Now no items show in the inventory for the pistol slot. 
Please hold on merging this for now as it seems it´s bugged. Current version in Dev25 is functional but just doesn´t allow new items to be added, so can be left alone.

Schizosaccaromyces:
Update to prevent multiple pistols being equipped (issue 1006), if it works.

pledbrook:
Added via [this commit](https://github.com/long-war-2/lwotc/commit/1765458039ccf5cc87dd333472475288c2035c75).

PR #1281, opened 2021-04-02
Allow items specified as Pistol Slot to show in Armory UI for soldiers

Schizosaccaromyces:
Changed the check for inventory availability so it allows pistols in  The List.


pledbrook:
Added via [this commit](https://github.com/long-war-2/lwotc/commit/1765458039ccf5cc87dd333472475288c2035c75).

PR #1289, opened 2021-04-16
Allow AI to consider overwatch counters for suppression targeting

NilchEi:
Saw MEC choose to suppress unit with lightning reflex, immediately tracked down the behaviours responsible for suppression targeting and added overwatch counter skill check.
TargetHasOverwatchCounter is modification of pre-existing HasOverwatchCounter behaviour.


NilchEi:
Closed because the change have minimum impact at current design

PR #1295, opened 2021-05-01
Grobo/chosen tweaks

Grobobobo:
Patch notes on discord


LongWarOfTheChosen/Localization/XComGame.int
```diff
@@ -6587,10 +6587,24 @@ LocLongDescription="Attack an adjacent target with your sword. Uses one action a
...
+LocFlyOverText="Bladestorm"
+LocPromotionPopupText="<Bullet/> If an enemy begins their turn in an adjacent tile, Bladestorm will trigger if that enemy tries to attack the <Ability:ClassName/>.<br/><Bullet/> If an enemy does not begin their turn in an adjacent tile, then Bladestorm will trigger when that enemy moves into melee range.<br/>"
+
+[AssassinBladestormAttack X2AbilityTemplate]
+LocFriendlyName="Bladestorm"
+LocLongDescription="Free <Ability:SecondaryWeaponName/> attacks on any enemies that enter or attack from melee range."
+LocHelpText="Free <Ability:SecondaryWeaponName/> attacks on any enemies that enter or attack from melee range."
+LocFlyOverText="Bladestorm"
+LocPromotionPopupText="<Bullet/> If an enemy begins their turn in an adjacent tile, Bladestorm will trigger if that enemy tries to attack the <Ability:ClassName/>.<br/><Bullet/> If an enemy does not begin their turn in an adjacent tile, then Bladestorm will trigger when that enemy moves into melee range.<br/>"
+
```
pledbrook, line 6607: This should be inside its own translation block (since if the current block has been translated already, a translator may miss this change).

pledbrook (changes requested):
Please improve the commit messages. They're currently almost meaningless to me, especially as they make a presumption that the reader already realises they're related to the Chosen, which is not a given once merged into `master`.

More importantly, it would be good to know what the main changes are about. The AI appears to have been changed so that the Chosen (or just the Assassin?) will ignore hazards when moving. If that's the case, it should be mentioned.

It's also worth mentioning why a new AssassinBladestorm ability was added instead of using the base version. I would even mention that Banzai no longer removes burning and acid (I assume that's the change) and why.

pledbrook:
Merged in [this commit](https://github.com/long-war-2/lwotc/commit/b39cf6a63de206c9b00202a233ac7cf6f709d115) and the one previous to it.

PR #1336, opened 2021-06-14
Add a Covert Action to obtain a Resistance MEC

K1ruka:
I felt like we needed some more early covert actions to populate the pool, and Resistance MECs are neat, so why not?

### DIFF
--- a/LongWarOfTheChosen/Localization/XComGame.int
+++ b/LongWarOfTheChosen/Localization/XComGame.int
@@ -2168,8 +2168,9 @@ strCharacterName="Resistance MEC"
 [Reward_Rebel X2RewardTemplate]
 DisplayName="Resistance Personnel"
 
-[Reward_ResistanceMEC X2RewardTemplate]
+[Reward_ResistanceMec X2RewardTemplate]
 DisplayName="Resistance MEC"
+RewardDetails="You'll get a Resistance MEC to guard one of your Havens."
 
 [Reward_NewResources X2RewardTemplate]
 DisplayName="New Resources"
@@ -2205,10 +2206,6 @@ RewardDetails="You'll get a selection of random enemy corpses."
 
 [Reward_Dummy_StatBoost X2RewardTemplate]
 DisplayName="Stat Boosts"
-
-[Reward_ResistanceMec X2RewardTemplate]
-DisplayName="Resistance MEC"
-RewardDetails="You'll get a Resistance MEC to guard one of your Havens."
 ; End Translation
 
 ; POI Templates
--- a/LongWarOfTheChosen/Localization/XComGame.int
+++ b/LongWarOfTheChosen/Localization/XComGame.int
@@ -2168,7 +2168,7 @@ strCharacterName="Resistance MEC"
 [Reward_Rebel X2RewardTemplate]
 DisplayName="Resistance Personnel"
 
-[Reward_ResistanceMec X2RewardTemplate]
+[Reward_ResistanceMEC X2RewardTemplate]
 DisplayName="Resistance MEC"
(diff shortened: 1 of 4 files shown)


LongWarOfTheChosen/Localization/XComGame.int
```diff
@@ -2205,6 +2205,10 @@ RewardDetails="You'll get a selection of random enemy corpses."
 
 [Reward_Dummy_StatBoost X2RewardTemplate]
 DisplayName="Stat Boosts"
+
+[Reward_ResistanceMec X2RewardTemplate]
+DisplayName="Resistance MEC"
+RewardDetails="You'll get a Resistance MEC to guard one of your Havens."
```
pledbrook, line 2211: There's already a localisation entry for this reward. Add the reward details to that one please.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2StrategyElement_DefaultCovertActions_LW.uc
```diff
@@ -68,6 +69,31 @@ static function X2DataTemplate CreateIntenseTrainingTemplate()
...
+	Template.Narratives.AddItem('CovertActionNarrative_ResistanceMec_Skirmishers');
+	Template.Narratives.AddItem('CovertActionNarrative_ResistanceMec_Reapers');
+	Template.Narratives.AddItem('CovertActionNarrative_ResistanceMec_Templars');
+
+    // NOTE: Soldier slots configured in `X2LWCovertActionsModTemplate` for consistency and
+    // because some Firaxis developer keeps making functions private..... The template mod
+    // also adds the Failure risk.
+
+	Template.Risks.AddItem('CovertActionRisk_Ambush');
+	Template.Risks.AddItem('CovertActionRisk_SoldierCaptured');
```
pledbrook, line 90: You want to have a capture risk for this? I don't know if the reward is worth it. I don't remember any other "easy" covert actions having a capture risk.

pledbrook (changes requested):
Looks good. See inline comments for 1 definite change and a question about having a capture risk for this CA.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2StrategyElement_DefaultCovertActions_LW.uc
K1ruka: I couldn't remember myself. I'm okay with removing that.

LongWarOfTheChosen/Localization/XComGame.int
```diff
@@ -2168,8 +2168,9 @@ strCharacterName="Resistance MEC"
 [Reward_Rebel X2RewardTemplate]
 DisplayName="Resistance Personnel"
 
-[Reward_ResistanceMEC X2RewardTemplate]
+[Reward_ResistanceMec X2RewardTemplate]
```
pledbrook, line 2171: I think this is case insensitive, but I don't understand why you want to change this when the reward name is explicitly 'Reward_ResistanceMEC'.

pledbrook (commented):
See inline. I'm fairly sure the localisation header change of case is safe, but it's a bit confusing when the existing reward name is capitalised. Wondering why the change (and because I'm not 100% certain the headers are case insensitive).

LongWarOfTheChosen/Localization/XComGame.int
K1ruka: That was just leftover from me making the reward template when I wasn't even aware that it existed in the first place.

pledbrook (approved):

pledbrook:
Thanks for the contribution!

PR #1339, opened 2021-06-15
Switch to X2ModBuildCommon v1.1.0

robojumper:
See <https://github.com/X2CommunityCore/X2ModBuildCommon>.

Everything below `.scripts/X2ModBuildCommon` is simply merged from the above git repository. Lwotc's build is thus pinned against a specific version of X2ModBuildCommon and doesn't require any `git submodule` interactions.
The scripts *can* be edited if needed, but these changes should be upstreamed in the long term.

Under the hood, this is more or less the same thing as before -- a PowerShell script invoked by a task configuration. The [top-layer script invoked](https://github.com/robojumper/lwotc/blob/x2mbc-switch/.scripts/build.ps1) is very simple and can be edited to add specific configuration (like cooking configuration) to the build script.

The only functional difference here is that it doesn't perform a full rebuild. A `clean` task is provided to allow a full rebuild if desired.


pledbrook:
Thanks for doing this robojumper. Much appreciated. One quick question that springs to mind after dealing with the FRA translations: is there a task for just converting and copying the localisation files to the installed mod? It would be useful for contributors to have something like that without needing the SDK and performing a full build.

robojumper:
Not in X2ModBuildCommon, and I don't think we'll be adding one considering X2ModBuildCommon generally assumes an SDK installation. Should be relatively trivial (~10 lines) to make a separate script for that though.

pledbrook:
Merged [here](https://github.com/long-war-2/lwotc/commit/087587d938af446b36f933b9f3f499cc0d09047c). Thanks!

PR #1369, opened 2021-07-06
Shieldbearer AI modification

NilchEi:
Fixes the problem of Shieldbearer AI always using shield on its first turn no matter what, while keeping its defensive behaviour intact; shielding 3+ units are still prioritized highest, but high priority shots are checked before TryEnergyShieldOnAnyTarget


pledbrook (approved):
Looks good, thanks!

PR #1374, opened 2021-07-07
Add a button on the Outpost Management screen to edit haven advisor loadout

ansabalirai:
Currently, the behavior when clicked via the avenger "Resistance Management" shortcut is a bit finnicky, in that backing out after editing the loadout results in two blank/unpopulated screens and then dumps you back in the barracks. I suspect the outpost management and resistance management screens are still in the stack but not showing for some reason. In any case, based on the recently submitted pull request (#1373), the player can access the resistance management from the geoscape and assign/outfit advisor without ever leaving the geoscape, so the need for this button when accessed through the avenger shortcut is not pressing. 
In any case, if the aforementioned jank is unacceptable, this button can be made visible only when viewed through the geoscape, where this issue does not occur...


pledbrook:
Merged in [this commit](https://github.com/long-war-2/lwotc/commit/4cd0a946cbba15cb2f91fe3461c256010d8bae79). Thanks!

PR #1379, opened 2021-07-11
Make Alien Facilities controller compatible

kdm2k6:
Modify `UIStrategyMapItem_AlienNetworkComponent_LW` so that Alien Facilities are clickable with a controller.

**Description** : A while back, `UIStrategyMapItem_Mission_LW `was modified so that strategy map items could be 'clicked' and would interact appropriately with a controller. Unfortunately, I didn't realize that the alien facility had its own special strategy map item class, `UIStrategyMapItem_AlienNetworkComponent_LW`.

Consequently, I have transferred all of the changes I made to `UIStrategyMapItem_Mission_LW` into `UIStrategyMapItem_AlienNetworkComponent_LW`.

Here is a general discussion of the changes made from my commits for `UIStrategyMapItem_Mission_LW`

=========================

Modifies `UIStrategyMapItem_AlienNetworkComponent_LW `--> `InitMapItem()`

When using a controller, an infiltrating mission displays a help icon, `StrategyScanButton `--> `consoleHint`, to the left of the haven information button. The problem is that it overlaps the icon representing the infiltrating squad. Consequently, the best way to deal with it, is to null it.

=========================

Modifies : `UIStrategyMapItem_AlienNetworkComponent_LW `--> `UpdateFromGeoscapeEntity()`

If an infiltrating mission map item is selected, make sure the infiltration percentage string is black. If an infiltrating mission map item is not selected, make sure the infiltration percentage string is its normal blue. If this is not done, the text will 'disappear' into the background colour.

Additionally, within this function, update the map item tooltip information when appropriate.

=========================

Modifies `UIStrategyMapItem_AlienNetworkComponent_LW `--> `OnMouseEvent()` and `OnUnrealCommand()`.

Originally, when a controller user clicked on the A button while a map item mission was highlighted, it would open a normal mission screen even if the mission was currently undergoing infiltration. As a result, `OnUnrealCommand()` had to be extended so as to call the same code as found in `OnMouseEvent()` --> `FXS_L_MOUSE_UP`. This code calls `OpenInfiltrationMissionScreen()`, which pushes an infiltration screen onto the screen stack if infiltration is occurring.

=========================

Added : `UIStrategyMapItem_AlienNetworkComponent_LW `--> `GetToolTipText`, `UpdateTooltip`, and `GenerateTooltip`

Provides the opportunity for controller users to display mission tooltips next to the mission map item. Formerly, they would only be accessible to mouse and keyboard users via the mission icons on the bottom icon bar.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIStrategyMapItem_AlienNetworkComponent_LW.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIStrategyMapItem_AlienNetworkComponent_LW.uc
@@ -6,10 +6,13 @@
 //           up when you sent a squad to infiltrate an alien facility with the base implentation of
 //           the parent class.
 // 
-//           This class' code is a literal copy paste of the UIStrategyMapItem_Mission_LW classs so
+//           This class' code is a literal copy paste of the UIStrategyMapItem_Mission_LW class so
 //           that the alien facility strategy item has the same infiltration mechanics as other missions
 //           do. All of the code in this class is unique to this class and does not override any of the
 //           parent class behavior.
+//
+//	KDM : The same controller code found in UIStrategyMapItem_Mission_LW has been implemented here so
+//	that alien facilities are 'clickable'. 
 //--------------------------------------------------------------------------------------- 
 
 class UIStrategyMapItem_AlienNetworkComponent_LW extends UIStrategyMapItem_AlienNetworkComponent;
@@ -40,7 +43,7 @@ simulated function UIStrategyMapItem InitMapItem(out XComGameState_GeoscapeEntit
 	InitPanel(Name(Entity.GetUIWidgetName()), 'MI_alienFacility');
 
 	PinImage = Entity.GetUIPinImagePath();
-	if( PinImage != "" )
+	if (PinImage != "")
 	{
 		SetImage(PinImage);
 	}
@@ -69,6 +72,13 @@ simulated function UIStrategyMapItem InitMapItem(out XComGameState_GeoscapeEntit
 	bScanButtonResized = false;
 
 	InstanceMapItem3DMaterial();
+
+	// KDM : The mission map item's help icon gets in the way of the infiltrating squad icon; the best way to deal with it is to null it.
+	// For more information on a similar topic please see the comments in UIStrategyMapItem_Region_LW --> InitMapItem().
+	if (`ISCONTROLLERACTIVE)
+	{
+		ScanButton.mc.SetNull("consoleHint");
+	}
 
 	return self;
 }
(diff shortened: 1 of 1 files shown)


LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIStrategyMapItem_AlienNetworkComponent_LW.uc
```diff
@@ -26,6 +29,7 @@ var UIProgressBar ProgressBar; // used to show remaining time for missions not y
 var int InDepth; // tracks depth of mouse in and outs, since it's recording from two possibly overlapping elements (2D and 3D icons)
 
 var localized string strRecon;
+var localized string InfiltrationTooltipString;
```
pledbrook, line 32: Doesn't this need an entry in the localisation files?

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIStrategyMapItem_AlienNetworkComponent_LW.uc
```diff
@@ -108,28 +119,45 @@ function InstanceMapItem3DMaterial()
...
-	local X2MissionTemplate MissionTemplate;
+	local int InfiltrationPct, InfiltrationTextColour;
 	local float ScanWidth;
+	local string InfiltrationPctValue, MissionInfo, MissionTitle;
+	local X2MissionTemplate MissionTemplate;
+	local XComGameState_LWPersistentSquad InfiltratingSquad;
+	local XComGameState_MissionSite MissionState;
 
-	if( !bIsInited ) return; 
+	if (!bIsInited )
```
pledbrook, line 129:
```suggestion
	if (!bIsInited)
```

pledbrook (changes requested):
Looks good, but pretty sure there's a missing l10n entry.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIStrategyMapItem_AlienNetworkComponent_LW.uc
kdm2k6: Correct.

If I already have it defined within LW_Overhaul.int as :
[UIStrategyMapItem_Mission_LW]
InfiltrationTooltipString="INFILTRATION"

Is there any way to pull the value from here, as in UIStrategyMapItem_Mission_LW, or do I need to make a separate entry :
[UIStrategyMapItem_AlienNetworkComponent_LW]
InfiltrationTooltipString="INFILTRATION"

pledbrook: You need to add an extra entry. The other option is to just pull the localised string from the other class. It's a public `default` value after all.

kdm2k6: So I'm assuming :
class'UIStrategyMapItem_Mission_LW'.default.InfiltrationTooltipString

pledbrook: Yes

LongWarOfTheChosen/Src/LW_Overhaul/Classes/UIStrategyMapItem_AlienNetworkComponent_LW.uc
```diff
@@ -70,6 +73,13 @@ simulated function UIStrategyMapItem InitMapItem(out XComGameState_GeoscapeEntit
 
 	InstanceMapItem3DMaterial();
 
+	// KDM : The mission map item's help icon gets in the way of the infiltrating squad icon; the best way to deal with it is to null it.
+	// For more information on a similar topic please see the comments in UIStrategyMapItem_Region_LW --> InitMapItem().
+	if (`ISCONTROLLERACTIVE)
+	{
+		ScanButton.mc.SetNull("consoleHint");
```
pledbrook, line 80:
```suggestion
		ScanButton.MC.SetNull("consoleHint");

```

pledbrook:
Thanks!

PR #1411, opened 2021-08-15
Fixing Repression activity

GafrieGithub:
As reminder of what repression is, this is the description from XComLW_Activities.ini: "Repression works only in uncontacted regions; it is simulating the Resistance without XCOM's assistance. It rolls several d100s once the cooldown is over."

As far as I can tell, and as far as I can remember this activity never worked, even in the original LW2. The reason for this is that RepressionComplete function seeks for RegionState only in the NewGameState, and immediately fails if it doesn't find it. It is unlike all other activities which seek XCOMHistory in such case. I am fuzzy on the details for why activities try to find RegionState object in NewGameState at all, since apparently it always fails, but simply doing what other activities do should work for Repression as well.
The second problem was that the iRoll variable simply wasn't assigned value properly, thus it would've had value of 0 for all checks.
Third problem was the condition check for REPRESSION_REBEL_LOST_CHANCE was inverted for reasons unknown. Probably a typo.

With this fixes this activity seem to work as seemingly intended. However, I should reiterate that I think it never ever worked, and thus might had never been playtested, balance-wise. Therefore, I am not completely certain that this fix should even be made, though I am leaning to it, since I like the idea behind the mechanic.
Looking at the chances of events happening, here is the average change of one repression activity: -0.02 Alert, -0.03 Vigilance*, =0 Rebels**, which should be multiplied by number of uncontacted regions to get average world-wide change per duration period.
(* slightly more, since vigilance decreases of REBEL_LOST events only happen if removed rebel was non-faceless, but it is hard to adjust for)
(** the percentage of faceless will rise over time, since events can add both faceless and normal rebels, but are removing only normal ones)

Since the Vigilance decrease of REPRESSION_REBEL_LOST and REPRESSION_2ND_REBEL_LOST events were not described as results in the ini file, I am not sure if Grobobobo had considered it when he made the last changes to the percentage chances on Nov 21, 2020. One might want to adjust the chances.


NilchEi:
Math is not my strongest point, but wouldn't average rebel change from vigilance activity actually be **+0.072 rebels**? 

rebel gain chance (20/100)(90/100)(90/100) = 0.162
rebel lost chance (80/100)(10/100)(90/100) + (20/100)(10/100)(10/100) = 0.074
2 rebel lost chance (80/100)(10/100)(10/100) = 0.008

0.162-0.074-0.008*2 = +0.072 rebels

So leaving regions uncontacted will outright favour XCOM in all counts.

NilchEi:
Repression setting that are closest to being absolutely fair would be this:
```
REPRESSION_RECRUIT_REBEL_CHANCE=13                   ; +1 rebel
REPRESSION_VIGILANCE_INCREASE_CHANCE=18                ; +1 vig
REPRESSION_CLONES_RELEASED_CHANCE=2                    ; +1 alert
```

REBEL 
(x/100)(90/100)(90/100) = (((100-x)/100)(10/100)(90/100) + (x/100)(10/100)(10/100) + 2((100-x)/100)(10/100)(10/100))
x = 12.087...
Rounded up to 13 to account for Faceless recruitments (+0.0083 rebel)



VIGILANCE
(2(2/100)(x/100)(90/100)(90/100)+(2/100)((100-x)/100)(90/100)(90/100)+(98/100)(x/100)(90/100)(90/100)+2(2/100)(x/100)(10/100)(90/100)) = 2(98/100)((100-x)/100)(10/100)(90/100)+(2/100)((100-x)/100)(10/100)(10/100)+(98/100)(x/100)(10/100)(10/100)+2(98/100)((100-x)/100)(10/100)(10/100)

(9 (225 + 113 x))/125000 = (24525 - 233 x)/125000
x = 18

Grobobobo:
I believe repression is *meant* to be overall favourable to xcom, like as a form of scaling

GafrieGithub:
> 
> 
> Math is not my strongest point, but wouldn't average rebel change from vigilance activity actually be **+0.072 rebels**?
> 
> rebel gain chance (20/100)(90/100)(90/100) = 0.162
> rebel lost chance (80/100)(10/100)(90/100) + (20/100)(10/100)(10/100) = 0.074
> 2 rebel lost chance (80/100)(10/100)(10/100) = 0.008
> 
> 0.162-0.074-0.008*2 = +0.072 rebels
> 
> So leaving regions uncontacted will outright favour XCOM in all counts.

These events are independent from each other, therefore 20% chance to gain the rebel is simply that, +0.2 rebels on average per repression. Similarly, REBEL_LOST is -0.1 rebel on average an another -0.1 for 2ND_REBEL_LOST.

NilchEi:
> These events are independent from each other, therefore 20% chance to gain the rebel is simply that, +0.2 rebels on average per repression. Similarly, REBEL_LOST is -0.1 rebel on average an another -0.1 for 2ND_REBEL_LOST.

Ah, are the timer for these repression events different? I assumed they are all rolled at the same time, so the region has 20% chance of gaining rebels, but chance of region gaining rebel *after* all repression rolls are lower because it has to avoid all the rebel loss events.

> I believe repression is meant to be overall favourable to xcom, like as a form of scaling, at least that's how I interpreted pavonis comments

Hmm maybe, the original values does reduce overall alert and increase rebel numbers.
However original Pavonis value for REPRESSION_VIGILANCE_INCREASE_CHANCE was 25 meaning overall vigilance was _increasing_ in response to aforementioned two XCOM-favoured changes. This balance was broken when you nerfed it's value to 15. Honestly I don't mind using the original value set by Pavonis team.

```
REPRESSION_ADVENT_LOSS_CHANCE=2                        ; -1 Alert, +1 Vig
REPRESSION_RECRUIT_REBEL_CHANCE=15                    ; +1 rebel
REPRESSION_VIGILANCE_INCREASE_CHANCE=25                ; +1 vig
REPRESSION_REBEL_LOST_CHANCE=10                        ; -1 nonfaceless rebel, min of 1, -1 vig
REPRESSION_CLONES_RELEASED_CHANCE=0                    ; +1 alert
REPRESSION_2ND_REBEL_LOST_CHANCE=10                    ; -1 nonfaceless rebel, min of 1, -1 vig
```

NilchEi:
I remember saying this looked unfavourable for XCOM, but it's -0.02 alert +0.0265 rebel +0.03 vig per repression event.

GafrieGithub:
> Ah, are the timer for these repression events different? I assumed they are all rolled at the same time, so the region has 20% chance of gaining rebels, but chance of region gaining rebel _after_ all repression rolls are lower exactly because these events are independent from each other, meaning rebel gain chance is 16% instead of 20% because it has to avoid all the rebel loss events as well.

No, the event is the same. All rolls are made within the same event, independent of each other. So, the chance 20% chance is simply the chance of this particular event triggering. What happens afterwards is irrelevant, and this  particular event results in net +0.2 rebel gain regardless. Which is why your 16% number of increasing rebel count after all events is true, but is also not the point at all.
If you want to count the net gain by listing all the outcomes and the outcomes  chances, then it is also fine, but for there "rebel lost chance (80/100)(10/100)(90/100) + (20/100)(10/100)(10/100) = 0.074" you forgot (80/100)(90/100)(10/100) outcome, which is different from (80/100)(10/100)(90/100) and adds another 0.072 to thet line sum, bringing total to 0.

NilchEi:
Never mind, my mistake. I left out one possibility on the equation so correct one is
(x/100)(90/100)(90/100) - 2((100-x)/100)(90/100)(10/100) + (x/100)(10/100)(10/100) + 2((100-x)/100)(10/100)(10/100)) = 0
so x=20 is indeed completely fair.

pledbrook:
I'll keep this open for review after 1.0, but it doesn't seem worth making the change before then since the strategy layer will have been balanced (probably unintentionally) around the non-functioning of this feature.

pledbrook:
Merged in above commit with some additional changes.

Issue #686, opened 2019-10-14
Restrictions from failed retals should clear after Liberation


TheReigningKing:
If you fail a retal, you're banned from performing the corresponding job action for 30(?) days.
If you liberate the region, you should be unbanned b/c the region is liberated now.

copyrite:
Addressed in #1413.

PR #1413, opened 2021-08-16
Clear job prohibitions after liberation

GafrieGithub:
This clears job prohibitions and retributions affecting the region when the region is liberated. The ingame notification may occur slightly after, but you can start assigning rebels to the job immediately.
Was reported in #686.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2StrategyElement_DefaultAlienActivities.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2StrategyElement_DefaultAlienActivities.uc
@@ -597,8 +597,7 @@ static function OnProtectRegionActivityComplete(bool bAlienSuccess, XComGameStat
 
 		//Removing prohibitions on jobs
 		Outpost = `LWOUTPOSTMGR.GetOutpostForRegion(PrimaryRegionState);
-		Outpost = XComGameState_LWOutpost(NewGameState.CreateStateObject(class'XComGameState_LWOutpost', Outpost.ObjectID));
-		NewGameState.AddStateObject(Outpost);
+		Outpost = XComGameState_LWOutpost(NewGameState.ModifyStateObject(class'XComGameState_LWOutpost', Outpost.ObjectID));
 		for (k = 0; k < Outpost.ProhibitedJobs.Length; k++)
 		{
 			Outpost.ProhibitedJobs[k].DaysLeft = 0;
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2StrategyElement_DefaultAlienActivities.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2StrategyElement_DefaultAlienActivities.uc
@@ -606,7 +606,7 @@ static function OnProtectRegionActivityComplete(bool bAlienSuccess, XComGameStat
 
 		for (k = 0; k < Outpost.CurrentRetributions.Length; k++)
 		{
-			OutPost.CurrentRetributions[k].DaysLeft = 0;
+			Outpost.CurrentRetributions[k].DaysLeft = 0;
 		}		
 
 		//distribute the RemainderAlertLevel amongst adjacent regions that haven't been liberated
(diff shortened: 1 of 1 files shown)


LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2StrategyElement_DefaultAlienActivities.uc
```diff
@@ -593,6 +595,20 @@ static function OnProtectRegionActivityComplete(bool bAlienSuccess, XComGameStat
 		RemainderAlertLevel = Max (0, PrimaryRegionalAI.LocalAlertLevel - AlertLevelsKilled);
 		PrimaryRegionalAI.LocalAlertLevel = 1;
 
+		//Removing prohibitions on jobs
+		Outpost = `LWOUTPOSTMGR.GetOutpostForRegion(PrimaryRegionState);
+		Outpost = XComGameState_LWOutpost(NewGameState.CreateStateObject(class'XComGameState_LWOutpost', Outpost.ObjectID));
+		NewGameState.AddStateObject(Outpost);
```
pledbrook, line 601: This should use the new WOTC API for game state objects:

```suggestion
		Outpost = XComGameState_LWOutpost(NewGameState.ModifyStateObject(class'XComGameState_LWOutpost', Outpost.ObjectID));
```

LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2StrategyElement_DefaultAlienActivities.uc
```diff
@@ -593,6 +595,20 @@ static function OnProtectRegionActivityComplete(bool bAlienSuccess, XComGameStat
...
+		for (k = 0; k < Outpost.ProhibitedJobs.Length; k++)
+		{
+			OutPost.ProhibitedJobs[k].DaysLeft = 0;
```
pledbrook, line 604:
```suggestion
			Outpost.ProhibitedJobs[k].DaysLeft = 0;
```

LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2StrategyElement_DefaultAlienActivities.uc
```diff
@@ -593,6 +595,20 @@ static function OnProtectRegionActivityComplete(bool bAlienSuccess, XComGameStat
...
+		}
+
+		for (k = 0; k < Outpost.CurrentRetributions.Length; k++)
+		{
+			OutPost.CurrentRetributions[k].DaysLeft = 0;
```
pledbrook, line 609:
```suggestion
			Outpost.CurrentRetributions[k].DaysLeft = 0;
```

pledbrook (changes requested):
Thanks for the fix. One change needed, but also a question: have you tested this? Just want to make sure. I think it should work, but that's not a reliable gauge. I'm just a little unsure that the outpost game state is reliably updated shortly after the alien activity is completed.

GafrieGithub:
> 
> 
> Thanks for the fix. One change needed, but also a question: have you tested this? Just want to make sure. I think it should work, but that's not a reliable gauge. I'm just a little unsure that the outpost game state is reliably updated shortly after the alien activity is completed.

I did test this as such: after liberation I could immediately assign rebels on the previously forbidden job. The notification of prohibition clearing appeared few hours later.

Regarding changing CreateStateObject to ModifyStateObject, I admit that I don't know full workings of this system, but I want to point out that the code elsewhere nearly always uses CreateStateObject when outpost is updated in similar manner. I don't know if it matters or not.

pledbrook:
> but I want to point out that the code elsewhere nearly always uses CreateStateObject when outpost is updated in similar manner

A fair point. This is because that code is LW2, which only had access to the original XCOM 2 API. Any new code we add should use `CreateNewStateObject()` and `ModifyStateObject()`. It's mostly for clarity and to remove the need for the extra `AddStateObject()` call.

pledbrook (approved):

pledbrook:
Thanks.

PR #1414, opened 2021-08-16
Fix to Chosen selecting uncontacted regions for Retribution

GafrieGithub:
Currently Chosen can sometimes select uncontacted region as a target for their Retribution activity.
I am fairly sure that if you iterate over array in such order, while removing  elements from it, you will skip over some regions and thus allow some uncontacted regions into the deck. Iterating in inverse order should fix it.


pledbrook (approved):
Good catch, thanks.

PR #1415, opened 2021-08-16
Remove Indomitable and Terrorize from templar's Ghosts

GafrieGithub:
No one bothered to add these two abilities to the exclusion list yet, so I might as well.
Indomitable should be removed because ghosts should not be able to raise their focus.
Terrorize should be removed because ghosts don't have Volt in the first place and so it is cleaner this way.

### DIFF
--- a/LongWarOfTheChosen/Config/XComGameData_SoldierSkills.ini
+++ b/LongWarOfTheChosen/Config/XComGameData_SoldierSkills.ini
@@ -172,6 +172,8 @@ STUNSTRIKE_COOLDOWN=2
 +AbilitiesGhostCantHave="GreaterPadding_CV"
 +AbilitiesGhostCantHave="GreaterPadding_MG"
 +AbilitiesGhostCantHave="GreaterPadding_BM"
++AbilitiesGhostCantHave="Indomitable"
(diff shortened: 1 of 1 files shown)


pledbrook (approved):
Thanks for the fix, but please stop commenting along the lines of "no one bothered to". That implies laziness rather than the more typical reasons of oversight, forgetfulness, or something else.

GafrieGithub:
> 
> 
> Thanks for the fix, but please stop commenting along the lines of "no one bothered to". That implies laziness rather than the more typical reasons of oversight, forgetfulness, or something else.
Well, sorry about that, but I cannot in good conciousness see it as anything else. It would probably take couple of minutes to fix it for Grobobobo or yourself, and I know for a fact that you were both aware of this. Yet it remained as it is for two months at least. I do not imply that any of you have a duty to do it or anything of the sort, but my impression was that it was too small of an issue to fix and it seemed better to spend time elsewhere, and this impression remains unchanged.

Grobobobo:
I think it's lazyness for terrorize and oversight for indomitable (I thought I removed that myself actually, did it get mixed up in merge again?)

PR #1440, opened 2021-09-21
Replace XComLWTuple with LWTuple in OnShouldShowPsi

GafrieGithub:
As far as I can understand OnShouldShowPsi function exists to provide support for Detailed Soldiers List mod (which I use), where the 'DSLShouldShowPsi' event is triggered in the first place. However, right now this function spam the logs with 'Warning: OnShouldShowPsi event triggered with invalid event data.' messages. It seems to me that it is because the DSL sends LWTuple-s instead of XComLWTuple objects. Changing it does fix it.

Visibly, this change replaced the display of ComInt stat for rookies with psi offense values. It may or may not be desirable, but it is what it was meant to do in the first place I think.

I don't know why OnShouldShowPsi expected XComLWTuple in the first place. Was it changed in DSL at some point? I am fuzzy on the distinction in usage of XComLWTuple / LWTuple. Given that there is some history with integrating/de-integrating DSL, I am also unclear on whether this function is even still meant to be there. In any way, even if it is just a remnant, then it should be removed altogether to not spam the logs.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2EventListener_Soldiers.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2EventListener_Soldiers.uc
@@ -367,10 +367,10 @@ static function EventListenerReturn OnOverridePersonnelStatusTime(Object EventDa
 
 static function EventListenerReturn OnShouldShowPsi(Object EventData, Object EventSource, XComGameState NewGameState, Name InEventID, Object CallbackData)
 {
-	local XComLWTuple			Tuple;
+	local LWTuple				Tuple;
 	local XComGameState_Unit	UnitState;
 
-	Tuple = XComLWTuple(EventData);
+	Tuple = LWTuple(EventData);
 	if (Tuple == none)
 	{
 		`REDSCREEN("OnShouldShowPsi event triggered with invalid event data.");


robojumper:
Totally unsolicited comment to clear a thing up:

> I am fuzzy on the distinction in usage of XComLWTuple / LWTuple.

`LW_Tuple.LWTuple` is the type that mods use to exchange structured information like argument lists through a single `Object` type to conform to `X2EventManager`'s interface. `LWTuple` comes from the `LW_Tuple` package, which every mod that makes use of it compiles and ships separately (so the definition cannot ever be modified from the definition we've retained since the vanilla Long War Toolbox (?)).

`XComGame.XComLWTuple` is the type the Highlander (replacing `XComGame` packages etc.) uses to exchange information with mods that depend on it. The reason for a separate type is that there's no way to compile the `LW_Tuple` package before `XComGame` in such a way that `XComGame` can depend on it, so it's an entirely separate type (the fact that it's shipped in `XComGame` allowed us to evolve the type to hold more complex types, like certain structs and arrays, see [Highlander docs](https://x2communitycore.github.io/X2WOTCCommunityHighlander/misc/XComLWTuple/)).

Thus the Highlander must use `XComLWTuple` for its events (and mods that depend on these events must cast the argument to `XComLWTuple`), and mods may choose to use `XComLWTuple` or `LWTuple` for their events (and the listeners on the other side must follow that choice.

Personally I would recommend that only Highlander events use `XComLWTuple` and that mods use `LWTuple` for their events, even if they depend on the Highlander for other reasons. This is what current DSL does, but the commit that introduced the integrated DSL used `XComLWTuple`: 16eaab4790ce9bae670455fdab115c74e0eb6e96. Either DSL was changed later, or @pledbrook performed an `LWTuple` -> `XComLWTuple` change?

Grobobobo:
I'm hesistant to merge because of 

> Visibly, this change replaced the display of ComInt stat for rookies with psi offense values.

We'll probably have to discuss on what it should actually show.

pledbrook:
> Personally I would recommend that only Highlander events use XComLWTuple and that mods use LWTuple for their events, even if they depend on the Highlander for other reasons.

I think my reasoning was that if you were going to depend on LWOTC, you will have the Community Highlander available anyway. But I realise now that this approach forces CHL as a dependency on other mods, so `LWTuple` should probably be preferred. On the other hand, `XComLWTuple` has more fields that can make for a much simpler API.

pledbrook:
Sorry, this should have been merged pre-1.0. I'll put it on `master`, but I don't think it's sufficient on its own to warrant a new version on Steam. If I do a 1.0.2 with other fixes too, then it will be available then.

pledbrook:
Implemented [here](https://github.com/long-war-2/lwotc/commit/23160be11c09163600609e9cb234cff9887535ed).

Issue #1447, opened 2021-09-28
Hyper-reactive pupils activates even when others miss a shot on the Ranger's target


Alchete:
v28.2

My Ranger had a 92% chance to hit on a drone.  Instead of firing with the Ranger, I took the shot with my Gunner and he missed.  

Back to the Ranger, and he now has +10 from Hyper-reactive Pupils and a 100% hit chance on the same drone.  He still has both actions.

Save game attached with Ranger having both actions and +10 from Hyper-reactive on *both* Drones within view.

EDIT:  I need more testing, but it honestly looks like if anyone misses anything, then Hyper-reactive activates.

[save_IRONMAN- Campaign 71.zip](https://github.com/long-war-2/lwotc/files/7240161/save_IRONMAN-.Campaign.71.zip)
e.

pledbrook:
Fixed with [this commit](https://github.com/GafrieGithub/lwotc/commit/6c53c8429010e451d15dba9fe04e805859f03ab8).

PR #1448, opened 2021-09-30
Fix Hyper Reactive Pupils PCS

GafrieGithub:
As described in #1447 HRP bonus currently applies after any unit misses, instead of just unit with HRP.
Now, technically speaking, one _could_ read the description such that "a miss"  in the HRP description refers to any miss whatsoever. And the fact that no check that the "miss was made by the unit with the PCS" was even attempted in the code _could_ mean that it was intended this way. However, I think it would be very bizarre thematically, and I believe no player ever thought that it would be how it works after reading its description.

Thus, the check for the miss to be made by the unit with HRP pcs is added.

Additionally, I added several entries to the HRP-related SHOTFIRED_ABILITYNAMES list, which do probably belong there. There are probably other fitting abilities missing form this list, but I couldn't think of any immediately.


bazik123:
my vote would be do NOT merge this, but instead to advertise everywhere Hyper Reactive Pupils arent as useless as they seem

PR #1449, opened 2021-09-30
Make Locked On not lose the bonus on non-offensive abilities

GafrieGithub:
Currently ranger's Locked On effect is lost when using any non-offensive ability or using non-primary weapon. This behavior differs from original LW2 and, I think, feels very janky.

From what I can tell, here is how it happened. Originally Locked On relied on XComGameState_Effect_LastShotDetails class to keep track of who was shot last. However, this class' event listener does not make the check that the unit using an ability is the same as the unit which have relevant effect (which is unclear as to why for me, but I suspect it is either the difference in how it is triggered in vanilla or how it was subscribed to in LW2).  Which lead to there being an issue where locked on was cleared by other soldiers using their abilities. That issue was fixed there #621 by replacing Locked On listener to vanilla ZeroIn listener.  However, said ZeroIn listener differs from the original RecordShot in certain details. Namely, it resets its counters if the unit uses any non-offensive ability or uses non-primary weapon.

The fix is to use custom listener, which mostly copied from ZeroIn, but with resetting counters removed.
Another difference from RecordShot is that it used SHOTFIRED_ABILITYNAMES whitelist of abilities which could apply Locked On. Now it uses ZeroIn criteria which is any targeted attack with primary weapon. I couldn't think of an ability which would be in one set but not another, but there could be some.

The second commit makes LockedOn both triggered by and applied to primary weapon shots only.  It is less obvious of a change to make, so take it  or leave it, but for the reasons outlined in said commit message I think it would be good change as well.

### DIFF
--- a/LongWarOfTheChosen/Config/XComClassData.ini
+++ b/LongWarOfTheChosen/Config/XComClassData.ini
@@ -1159,7 +1159,7 @@ NumInDeck=0
 							   (AbilityName="FieldMedic"), \\
 							   (AbilityName="PrecisionShot", ApplyToWeaponSlot=eInvSlot_PrimaryWeapon), \\
 							   (AbilityName="SmokeGrenade"), \\
-							   (AbilityName="LockedOn",  ApplyToWeaponSlot=eInvSlot_Unknown), \\
+							   (AbilityName="LockedOn",  ApplyToWeaponSlot=eInvSlot_PrimaryWeapon), \\
 							   (AbilityName="Flashbanger",  ApplyToWeaponSlot=eInvSlot_Unknown), \\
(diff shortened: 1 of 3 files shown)


Alchete:
Thanks Gafrie.  Will this also fix Locked On not persisting through to the next XCOM turn?  The current behavior also differs from LW2.  As it stands in 28.2, if the Ranger's first shot at a target is his second action or if it's a turn-ending shot like Walk Fire, the Ranger does not have the bonus the next turn.

GafrieGithub:
Right, I thought it would do so by default, but I forgot that the values set by using SetUnitFloatValue are set to be cleaned up every turn by default. I added another change and now it should persist through turns though.

PR #1454, opened 2021-10-04
Small change to rounds' PurePassive effects and small fix for stiletto rounds

GafrieGithub:
There are couple of small issues I noticed while working on Flechette Rounds.

First is that the PurePassive effects for every ammo rounds, which are used to display ammo description in soldiers F1 and to display an icon in lower-left, use the default 'eAbilitySource_Perk' value in their creation. Where I believe 'eAbilitySource_Item' would be more appropriate. Visually, it changes the color of the lower-left icon to blue-ish color.

Second issue is that Stiletto Rounds use custom X2Effect_StilettoRounds for its effect, but also an additional PurePassive effect Stiletto_Rounds_Ability_PP to show its description. Additionally, "Stiletto_Rounds_Ability" template, which creates X2Effect_StilettoRounds, lacks an entry in the loc file. Only the entry for Stiletto_Rounds_Ability_PP is there. In-game, it could be noticed in the damage preview window: one of the "+1 damage" entries for stiletto rounds (the conditional one, for which X2Effect_StilettoRounds is responsible) has blank text instead of a name.  It seems to me that it would be better there to just use the actual effect to show its description and not create an additional dummy one.


Grobobobo:
Try Fixing the merge conflict, if you can.

GafrieGithub:
Done.

PR #1477, opened 2021-11-14
Fix Paramedic visualization

GafrieGithub:
If a Reaper have Paramedic skill he will currently shoot in the air at the beginning of a mission. I believe assigning TypicalAbility_BuildVisualization is unneeded here.


Grobobobo:
Already fixed with paramedic rework

PR #1478, opened 2021-11-14
Fix Jammer charges persistence

GafrieGithub:
Currently, if you use Jammer ability, save the game, then reload it, Jammer charges will be restored.
Adding this line (taken from TypicalAbility_BuildGameState) fixes it.


Grobobobo:
I'll fix it with the paramedic rework.

PR #1479, opened 2021-11-14
Fix Focus Fire inconsistent applicability for SPARKS

GafrieGithub:
Currently, SPARKs' main weapon attacks get bonus aim from already applied stacks of Focus Fire, but do not contribute to it. Adding 'sparkrifle' to Focus Fire VALIDWEAPONCATEGORIES list make SPARKs fully benefit from Focus Fire.
It may be argued, that it should be made that the SPARKs do not get _any_ Focus Fire benefits, including the aim, due to how officer abilities generally do not apply to SPARKs (with couple exceptions). However, making both parts works is an easier fix to do, so I chose this approach.


ansient0:
It can be argued the status quo behavior is consistent with Sparks not acting on officer commands -- they don't pile on to focusing with their own rifle shots, but do benefit from anyone else having focused fire in a way that makes the target easier to hit

PR #1480, opened 2021-11-16
Fix MAX_FACELESS_PROPORTION check

GafrieGithub:
Since both GetNumFaceless() and GetRebelCount() function results are of "int" type the division applied is here is the integer division. So, GetNumFaceless() / GetRebelCount() can only be equal 1 if all rebels are faceless and 0 otherwise. Thus, this part of the check will always pass if there is at least one non-faceless rebel in the haven, and so MAX_FACELESS_PROPORTION being set to 0.32f is meaningless, as it could  have been set to 0.99f with the same effect.

Casting function results to float should fix it.

This change does affect strategic layer. Players could previously get havens with like 4/5 faceless, which would no longer be possible with this change.  As I understand there is certain hesitance to change strategic layer, so  I leave the decision to whether fix it or not to others discretion.


GafrieGithub:
As an afterthought, if the current behaviour is to be preserved, then it would better to still apply this fix, but change MAX_FACELESS_PROPORTION config value to 0.99f. Which I believe would be equivalent to what we have now. This way at least it would not be misleading later on.

Grobobobo:
Eh, I think 4/5 haven faceless is not a desirable feature of this mod.

PR #1514, opened 2021-12-24
Fix some code issues where uc-analyzer gives up

robojumper:
Apart from the dangling else fix, these should not affect the behavior at all, but allow my more restrictive [`uc-analyzer`](https://github.com/robojumper/uc-analyzer) to continue analysis instead of giving up with parser errors, type errors, or resolution errors.

---

The enum variants were found by a lint of the following form:
```
error[body-lowering-error]: type mismatch: expected enum<GameRuleStateChange>, found primitive int
   --> X2Ability_PerkPackAbilitySet:751:61
    |
751 |     WillBonus.BuildPersistentEffect (1, true, false, false, 7);
    |                                                             - 7 corresponds to eGameRule_PlayerTurnBegin
    |
```
so I'm confident I picked the correct variants, but this could be a good opportunity to double check those variants in particular.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2StrategyElement_LWObjectives.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2StrategyElement_LWObjectives.uc
@@ -262,7 +262,7 @@ static function CreateBlacksiteMission_LW(XComGameState NewGameState, XComGameSt
 	RewardTemplate = X2RewardTemplate(StratMgr.FindStrategyElementTemplate('Reward_None')); // no rewards for completing story objectives
 	RewardState = RewardTemplate.CreateInstanceFromTemplate(NewGameState);
 	Rewards.AddItem(RewardState);
-	MissionState = CreateMission(NewGameState, Rewards, 'MissionSource_BlackSite', 8); // as far away from player as possible, because reasons
+	MissionState = CreateMission(NewGameState, Rewards, 'MissionSource_BlackSite', eGameRule_PlayerTurnEnd); // as far away from player as possible, because reasons
 
 	RegionState = MissionState.GetWorldRegion();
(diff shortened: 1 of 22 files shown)


pledbrook (approved):
Did you do these manually?! Thanks, I'll get on it.

LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2StrategyElement_LWObjectives.uc
```diff
@@ -262,7 +262,7 @@ static function CreateBlacksiteMission_LW(XComGameState NewGameState, XComGameSt
 	RewardTemplate = X2RewardTemplate(StratMgr.FindStrategyElementTemplate('Reward_None')); // no rewards for completing story objectives
 	RewardState = RewardTemplate.CreateInstanceFromTemplate(NewGameState);
 	Rewards.AddItem(RewardState);
-	MissionState = CreateMission(NewGameState, Rewards, 'MissionSource_BlackSite', 8); // as far away from player as possible, because reasons
+	MissionState = CreateMission(NewGameState, Rewards, 'MissionSource_BlackSite', eGameRule_PlayerTurnEnd); // as far away from player as possible, because reasons
```
pledbrook, line 265: Oops 😄  I'm glad I just changed this code today, otherwise I wouldn't have noticed!

pledbrook:
Merged with [this commit](https://github.com/long-war-2/lwotc/commit/463b1df63bfe305b40629636b14211bde2c375a8). Thanks!

LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2StrategyElement_LWObjectives.uc
robojumper: Embarassing, sorry. That was me taking a Ctrl-F shortcut to find more of these per cycle because uc-analyzer stops after the first type error.

PR #1543, opened 2022-02-12
Update russian description for Death From Above

1rb1s1rb1s:
Previous russian description for DFA was a bit misleading. It stated that killing an enemy from higher elevation would only cost one action (while in fact it REFUNDS one action). This could lead one to think that DFA on reapers, for example, doesn't complement Light 'Em Up, because LEU already makes standard shots not end the turn, which would, according to old DFA description, mean that Reaper with DFA+LEU will be able to shoot twice a turn regardless of kills (while in fact killing an enemy enables Reaper to take three shots). This update makes it clear that action will be REFUNDED in case of a kill.


FlashVanMaster:
this one makes clear that not refunded, but giving extra action point, and i agree with that. (also ye, DFA description from lw2 is not correct)
Ну кста мог и дописать что работает только один раз за ход

PR #1593, opened 2023-06-13
Update LW_Tutorial.int

illogicalproductions:
Updating Chosen Tutorial


Tedster59:
small localization change, merging.

PR #1603, opened 2023-07-27
Update XComGame.rus

FlashVanMaster:
revert changes that were incorrect and made years ago


Tedster59:
Thanks for providing this.  I'll get it added.

PR #1633, opened 2023-12-18
Display Banish hit mod as int

copyrite:
After the most recent change to Banish loc, the aim penalty was displaying as `20.0000`.


copyrite:
Already in as 67bf486b10b5ce0f71e09c3f642f81933d148b5a

PR #1707, opened 2024-03-17
Replace Sidewinders' Hit and Slither with Reposition

copyrite:


Tedster59:
I'm not sure how I feel on this one, Hit and Slither feels like a fun bit of flavor even if it's technically no different from Reposition.

copyrite:
The main problem is that seeing the flyover easily leads one to draw a conclusion that it's a copy of Hit and Run.

PR #1714, opened 2024-03-20
Revert Whiplash CBAC bandaid fix

copyrite:
CBAC no longer ships with the config that forces Whiplash to be 0AP, we can revert our bandaid.


xgl1991714:
I have tested it, and it does work. Thx.

PR #1724, opened 2024-03-29
Fixed the issue where Crusader's Rage sometimes had lower damage bonu…

xgl1991714:
5fce9c05b1667724358e19c95b858fb7
9902ef675c6c82522fee101444a5a2a3
As shown in the images above, when the health is noticeably below 50%, the damage bonus of Crusader's Rage clearly does not reach 50%. Through analysis and testing, it was found that the damage bonus of Crusader's Rage is split into two parts when health is at or below 50%: one part is 25%, and the other part is 20%, resulting in a final effect of (1+0.25)*(1+0.2)=1.5. However, due to floating-point arithmetic and rounding, this splitting scheme sometimes led to instances of lower damage. To mitigate this, the two split parts are now merged, aiming to minimize occurrences of reduced damage.


Tedster59:
Was tested by Copyrite, merging.

Issue #1547, opened 2022-03-04
Invasion mission starts with 0 Rebels to save


attilabertok:
Playing version 1.0.1
I got an invasion mission, advent trying to take back a liberated region. There are two objectives, rescue as many rebels as possible, and destory the psionic relay.
The rescue counter shows 0/0 rebels to save (there are a number of rebels on the map, thoug), and no matter how many civilians I activate, as soon as I destory the relay, I get a "decapitation" faliure for the mission.

wdanny69:
this one will need to be looked further, do you have a save so we can try some things out.

also in invasion missions, not all civilians are actual rebel leaders which is why it did nothing for you.

PR #1731, opened 2024-04-06
add min haven count to invasion activity 

wdanny69:
This adds the min haven count condition to the invasion activity similar to haven retaliations so that there are still some rebels that can be rescued. This should help resolve [Invasions with 0 rebels](https://github.com/long-war-2/lwotc/issues/1547) that just ends up in an automatic failure as there was no one to rescue.

Currently just using the haven retaliation count as it made the most sense to keep the two the same. 

Also updated the logging for "MinRebel" to make it clear whether the activity can launch in a given region.


copyrite:
From a fluff perspective, this doesn't seem right to me. The aliens should _want_ to invade the region even when there are 0 rebels. Therefore I'd prefer if this would use a mechanism where the mission fails as soon as it spawns (if there is a way to implement that). I realize this is an easier way out to a somewhat rare situation, though.

wdanny69:
flavor wise while i agree that does make more sense,

from a gameplay perspective this is more an unfun mechanic and a very swingy one at that when there is very little counterplay to that. i know this is more covering a bit of an edge case as its not very often to reach a liberation status while also having the haven dead but its better this than to allow them to take it for basically free

PR #1738, opened 2024-04-13
Make Holotarget series skills silent

xgl1991714:
Added `Template.bSilentAbility = true` to three HoloTarget-related skills. This way, DoesAbilityCauseSound() will return false, and HoloTarget won't trigger noise-related calculations anymore.
Not sure if it will affect the Rulers' reaction.
QQ截图20240413185620


wdanny69:
is there a point to this, the holotargeter is already configured to have 0 sound range but that way it can always be configured if someone decides they want to

xgl1991714:
> is there a point to this, the holotargeter is already configured to have 0 sound range but that way it can always be configured if someone decides they want to

Previously, I've encountered numerous game crashes caused by Holotarget, with the logs always ending with 'Sniper Defense Weapon sound generated by XCom'. So, this change was made in hopes of reducing the frequency of crashes.

wdanny69:
ok, if this does help it then no reason not to do it either since the intended design was to be silent anyways

copyrite (approved):

PR #1745, opened 2024-05-01
Permanent Advent Firewalls Dark Event from 100% to 33% of mechanicals

SoupyDelicious:
(Go easy, it's my first time on Github xD) 

Request to change Advent Firewalls - Permanent Dark Event to affect only 33% of encountered robots, as seen on wiki.  

Secondary to the fact that it's not fun gameplay, it's too harsh.  Penalties for failed hacks is already very high.  

Perhaps a nerf to Failsafe would do instead, having it only prevent failed hack penalties a % of the time (75%, for example)


copyrite:
From a technical and process viewpoint, you did pretty well. This kind of a change would have to go through the `beta` branch, which I've changed for you. I'd also make the change on a feature branch, not my own `master`, but that's a problem for you, not us. I can give you directions over on Discord, if you like.

As for the change itself, I strongly disagree with it. I consider myself a heavy Specialist user and don't recognize this problem with hack defense. It's also very easy to just buff XCOM out of community suggestions so buffs should really be accompanied with a stronger argument than "not fun". I'd really like to throw first-time contributors a bone, but I unfortunately wouldn't want to see this change go through.

SoupyDelicious:
Hey there Copyrite, thanks for taking the time :D  

Ah, so on that 'feature branch' point, if I were to be doing a lot of my own modding, it'd be prudent to make my own mini-beta branch?  Is that effectively the purpose of that?  Appreciate you offering further help on discord.  I have nothing else for now, but this github thing is definitely very intuitive and cool.

Regarding the change: not that you could know, but I'm notorious for not only being annoying, but also for erring on the side of hardcore/unforgiving gameplay wherever there's an option. 

When I say 'fun', I guess I'm just referring to the kind of fun that isn't mutually exclusive to 'it benefits me, therefore it feels good' type-fun.   It just 'feels bad', as subjective a statement as that is.  RNG is fun, and to come up against an enemy who was 'lucky enough' to remain un-buffed is a real micro-win, dopamine moment.  

We could change it to 66% of enemies are effected by firewall.  Even a small chance of being spared; of the game throwing something in your favour is a more fun version of that mechanic.  And if Specialists are already powerful enough that such a widespread debuff is necessary, then there are more 'under-the-hood' ways to alleviate that, but that's a whole 'nother discussion =P

Tedster59:
I recommend instead of hardcoding the value, expose it to config instead, similar to how the others there are configurable.

PR #1783, opened 2024-12-15
Optimize some Chinese translations

xgl1991714:


Tedster59:
opps, missed this one.  will get it in.

PR #1798, opened 2025-02-09
Use the local value of MT_DEFENSE and MT_DODGE in X2Effect_MovingTarget

mrhyperventilate:
The recent fix to Moving Target accidentally set the defense and dodge values to 0.

MT_DEFENSE and MT_DODGE were changed from config variables to simple instance variables in 0ef85b2db, so the default property value of these variables is zero.


Tedster59:
Thanks!

PR #1803, opened 2025-03-19
Preserve HP and Will deficits in RespecSoldier

mrhyperventilate:
I've been using this handy function for testing out new abilities and class trees but noticed that it reset HP and Will. That broke the continuity of the campaign a bit, so I decided to add this change for convenience.

Happy to close if this is intended behavior.


Tedster59:
This is reasonable, since I don't think the command cleans up the heal/will projects, so it means soldiers can get "bricked" while at full HP and not be able to sent out.

Issue #1624, opened 2023-11-11
Game crashes when Warlock chosen is killed on retaliation mission


kaphula:
Game crashes on ironman when chosen is killed. Only LWOTC2 mod from Steam and its dependencies installed. OS: Linux.

On the save file you should be able to double barrel dead the chosen on the same turn: 

[save_IRONMAN- Campaign 546.zip](https://github.com/long-war-2/lwotc/files/13325062/save_IRONMAN-.Campaign.546.zip)

Tvrdoglavi:
I have the same issue on first time killing the chosen in the game. Form it was one of those "Unhindered Operations" missions.

nusendra:
I have alot of bad experiences on playing ironman. Im pretty sure somehwre said that they don't recommend ironman on LWOTC

richard-melvin:
on windows and latest beta, the provided save doesn't crash, either when the warlock dies, or the mission completes. nothing really unusual in the log. One soldier is missing a headshot (placeholder img instead).

Tvrdoglavi:
The timing of the crash is interesting. I assume that the game calculates the result of an attack before animating it. The game crashes the moment I click on the attack that would kill the chosen. It is the same with both of the runs where I have this crash. It makes it a game breaking bug since it is not possible to avoid the chosen.
I'm attaching my two saves. I hope that helps. 
[Save Files.zip](https://github.com/long-war-2/lwotc/files/13491687/Save.Files.zip)

Tvrdoglavi:
Yes, it's there is something about the large genelab map that crashes the game on Linux. That was a relatively easy fix to remove the map from the game. Unfortunately, removing the chosen is not really an option here. I've enjoyed the mod quite a bit until the point where it crashes. I hope to be able to play it at some point.

Tedster59:
do you have the log files from these crashes too?

kaphula:
Could you instruct me where they are located, if you know?

Tvrdoglavi:
My understanding is that .dmp is the only crash file created on Linux. No crash logs are generated.

mdbowler:
I have exactly the same issue with the first time killing a chosen on a retaliation mission. As soon as it dies, either by being shot or burning, the game crashes.

Platform: Linux, not Ironman

Dump and save attached.

Save: [lwotc_retaliation_crash.zip](https://github.com/long-war-2/lwotc/files/13805134/lwotc_retaliation_crash.zip)

Crash dump: 
[...]

PR #1806, opened 2025-03-21
Fix crash on Linux on killing a Chosen

mrhyperventilate:
Use of an uninitialized variable. We can make guesses about why this crashes on Linux and not on Windows, but this is a code error on Windows as well.

Fixes https://github.com/long-war-2/lwotc/issues/1624

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2Effect_ChosenLoot.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2Effect_ChosenLoot.uc
@@ -32,9 +32,15 @@ static function EventListenerReturn ScavengerAutoLoot(Object EventData, Object E
 		return ELR_NoInterrupt;
 	}
 
-    
     if(DeadUnit.IsChosen())	
     {
+		EffectState = XComGameState_Effect_EffectCounter(CallbackData);
+		if (EffectState == none)
+		{
+			`RedScreen("Chosen Loot Check: no effect game state");
+			return ELR_NoInterrupt;
+		}
+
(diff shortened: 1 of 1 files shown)


AV00:
Any idea when a new release containing this fix will be available ?

mrhyperventilate:
Steam decided to convert my XCOM 2 native Feral port into a Windows+Proton installation last night, so that might indicate that Valve is dropping support for the Linux version. I was annoyed at first, but after moving over my configuration I have to admit that running XCOM 2 with Proton is simply superior to the Feral version:

- No more random segfaults like this one due to runtime differences
- Slightly better performance
- Decreased memory usage (this was a pleasant surprise)
- Some broken textures from the workshop now render correctly
- Logging to Launch.log actually works as expected

I suspect that Valve will forcibly convert your installation as well. If you're not running in Steam, then disregard this message.

mrhyperventilate:
> I am running Steam so ... it should happen automatically next time I try to launch it ?

I don't know. I could be part of a test group, we could be in the middle of a rolling deployment, or it will definitely convert the next time you launch the game.

> And what do you mean by "moving over my configuration" ? You had to do some manual steps ?

The update moved my saved games, but I had to copy my Photobooth, Config, and CharacterPool directories into the new wineprefix. If you're not in the middle of a game, this would be a great time to reconfigure everything from scratch.

AV00:
Ok, I have the answer to all my questions: few hours ago Steam updated its subsystems completely (like you were saying) and indeed I have a "Windows version" now. So I was able to prove to the hunter chosen what buffoon he is.
It wasn't easy get everything set up again as the mods were disabled and enabling them took some attempts with both new and old mod launcher. But after a bit of fiddling everything is working.

I can confirm that the photos from the Photobooth are not found, I guess they're just around.

Thanks anyway.

Issue #1807, opened 2025-03-26
RewardTemplates being modified after OPTC for no good reson


Styrium:
Some of the template modifications made by LW_Toolbox run at OnLoadedSavedGameToStrategy because it requires some settings recorded to its XGS. The modification of RewardTemplates is implemented in a static function that doesn't refer to the XGS and should probably be moved to OPTC. I'll hopefully be moving it to the proper time so other mods can overwrite it with the proper CHL run order.

PR #1808, opened 2025-03-26
RewardTemplates being modified after OPTC

Styrium:
Some of the template modifications made by LW_Toolbox run at OnLoadedSavedGameToStrategy because it requires some settings recorded to its XGS. The modification of RewardTemplates is implemented in a static function that doesn't refer to the XGS and should probably be moved to OPTC. I'll hopefully be moving it to the proper time so other mods can overwrite it with the proper CHL run order. Fixes Issue #1807

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Toolbox_Integrated/Classes/X2DownloadableContentInfo_LWToolbox.uc
+++ b/LongWarOfTheChosen/Src/LW_Toolbox_Integrated/Classes/X2DownloadableContentInfo_LWToolbox.uc
@@ -94,8 +94,6 @@ static event OnLoadedSavedGameToStrategy()
 
 	ToolboxOptions.RegisterListeners();
 	ToolboxOptions.UpdateWeaponTemplates_RandomizedDamage();
-	//This is a static function, it should probably run at OPTC instead of at runtime - styr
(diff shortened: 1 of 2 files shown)


LongWarOfTheChosen/Src/LW_Toolbox_Integrated/Classes/X2DownloadableContentInfo_LWToolbox.uc
```diff
@@ -94,7 +94,8 @@ static event OnLoadedSavedGameToStrategy()
 
 	ToolboxOptions.RegisterListeners();
 	ToolboxOptions.UpdateWeaponTemplates_RandomizedDamage();
-	ToolboxOptions.UpdateRewardSoldierTemplates();
+	//This is a static function, it should probably run at OPTC instead of at runtime - styr
+	//ToolboxOptions.UpdateRewardSoldierTemplates();
```
copyrite, line 98: Likewise here, removing both commented lines suggested.

Styrium:
Since the other stuff at X2DLCI_LWToolbox is only called at OnLoadedSavedGameToStrategy, it still needs to be updated on a Tactical load; it currently handles this by a UISL tied to the UITacticalHUD screen. Is there literally any good reason not to just have it run at OnLoadedSavedGameToTactical?

PR #1816, opened 2025-04-22
Updated German Localization

NorDogroth:
This pull request updates and expands the German localization to match the current state of the mod. All .deu files should now contain the complete and up-to-date set of strings.

Additionally, the encoding of the files has been set to UTF-16 to ensure correct display of German special characters (e.g., umlauts).


copyrite:
> Additionally, the encoding of the files has been set to UTF-16 to ensure correct display of German special characters (e.g., umlauts).

Can you elaborate this? The preferred encoding for Git is UTF-8, and localization files are converted to UTF-16 build time.

NorDogroth:
Thanks for the clarification – I downloaded the mod via Steam Workshop and in that version the .deu files are still in UTF-8. This caused umlauts and special characters to display incorrectly in-game. After converting the files manually to UTF-16 LE, everything displayed correctly.

So it seems the build-time conversion to UTF-16 might not be happening in the published version?

copyrite:
Both me and the main dev are finding that the files shipped with the mod are little-endian UTF-16.

One source of uncertainty is the version. Do you have a specific workshop distribution of LWOTC in mind that you think is shipping in wrong encoding?

NorDogroth:
That's right, in the mod they are correctly delivered as UTF-16 LE.

It's good to know that the conversion happens properly during the build process. I've noticed that many other mods I translated were delivered in UTF-8, which sometimes causes issues.

I understand that for Git we should keep the files in UTF-8, and I've re-encoded the German files accordingly.
Everything should be fine now. Thanks for the clarification!

copyrite (approved):
Much better now.

PR #1955, opened 2026-05-05
Patch 1

ToallaNova-O2D:
Fixed some syntax and context errors on some of the translations in the second wave menu, to keep it as neutral as posible. Reworded some descriptions for better clarity.


ToallaNova-O2D:
Included some fixes for spanish translations, focused on syntax and context.

PR #1957, opened 2026-05-09
Refactor Wound Reduction to use UnitEndedTacticalPlay

Styrium:
refactored Greater Padding to use the basegame systems to handle the wound reduction. It should, hopefully, be less cursed than the current implementation


Styrium:
don't merge this yet
It only now occurred to me that, although I have tested that the wound reduction code doesn't run in-between the parts of a multi-part mission, I haven't tested if the effect is reapplied on subsequent parts. Gotta double-check that

Styrium:
double-checked. only one effect on the unit in the 2nd part of the mission.

PR #1960, opened 2026-05-23
Improve Justice and Wrath loc

copyrite:


ToallaNova-O2D:
Any updates here?

copyrite:
Don't pollute the thread with unnecesary remarks.

M3r1st:
I suggest using `SelfCooldown_LW` tag instead of the one provided by Firaxis because of the difference in logic.

copyrite:
Noted and changed, but didn't test.

Tedster59:
Ok, (A) it's still in draft, (B) now it's mad about merge conflicts

copyrite:
Draft is correct, I don't think a consensus was reached when this was last discussed. I'm not personally that invested in what the final text should be.

Might resolve conflicts later.

PR #1974, opened 2026-07-07
Update GREMLINs to affect Blinding Protocol

M3r1st:
1. Create and add missing passives to GREMLIN templates.
    - These passives match the ability names used for bonus radii in `X2AbilityMultiTarget_Radius` of Blinding Protocol.


ToallaNova-O2D:
Hi, all good here?

PR #1976, opened 2026-07-31
Update build instructions

Puppyguarb:
I recently built the mod from `master`, the current build instructions are missing a couple details.

First, the media assets has some files in the wrong location. Specifically `md_Warehouse_02.umap` and `sm_MegaBlock_Middle_Diner_LW.umap`, they are in `LongWarofTheChosen/Content/Parcels` rather than `LongWarofTheChosen/ContentForCook/Parcels` which causes a build fail. This could be fixed by updating the media assets file, but so long as we tell new devs it's probably fine.

Second, the instructions for the line replace in `XComGameState_Unit.uc` were accidentally a duplicate of the instructions for the line replace for `SeqAct_SpawnAdditionalObjective.uc`.

No AI was used in the creation of this pull request or its commits because I am not a fan of AI.


copyrite:
Good catch on the XCGS_Unit. I even asked people to eyeball my changes and no one stepped up.

Tedster59:
I can definitely update the instructions, I thought I added those specific parcels to the tracked list so gitignore didn't exclude them, after the mistake that was trying to use git-lfs once I don't want to deal with that again.

PR #1977, opened 2026-07-31
Added fix for reinforcement spawning freeze

Puppyguarb:
# The Issue
There is an issue that I encountered with reinforcement spawning sometimes causing the game to freeze, this is due to the while loop for finding a reinforcement spawn location iterating infinitely, which is caused because `SPAWNMGR.GetCurrentXComLocation()` can return a position that is out of bounds.

I have had this issue happen on several different missions, seemingly at random. Bear in mind I am also running well over 100 mods, so it's certainly possible this issue has nothing to do with LWOTC and is really only my problem. But in that case this should still be considered as at least a compatibility fix.

Here is the game save that was having this issue, feel free to use it for testing but again there are quite a few mods used in it: [game_save.zip](https://github.com/user-attachments/files/30601607/game_save.zip)
# The Fix
My solution was just to calculate XCom's average position manually rather than relying on `SPAWNMGR.GetCurrentXComLocation()`. This is done with a fittingly named function called `GetXComLocation()` added to `SeqAct_LWCallReinforcements.uc`.
# Additional Notes
The function `GetPositionForCivilian` in `LWTacticalMissionUnitSpawner.uc` also uses `GetCurrentXComLocation()` from the `SPAWNMGR`, which means it could potentially have a similar problem. I have not touched this file since I never pay attention to civilians in my games and have never noticed an issue.

No AI was used in the creation of this pull request or any of its commits because I am not a fan of AI.

### DIFF
--- a/LongWarOfTheChosen/Src/LW_Overhaul/Classes/SeqAct_LWCallReinforcements.uc
+++ b/LongWarOfTheChosen/Src/LW_Overhaul/Classes/SeqAct_LWCallReinforcements.uc
@@ -228,7 +228,7 @@ function SpawnReinforcements(int ReinforceIndex, XComGameState NewGameState)
 				// LWOTC: Since the reinforcement spawner doesn't seem to do a good job of
 				// randomisation spawn locations, we manually randomise the spawn location
 				// and use a smaller tile offset for InitiateReinforcements().
-				Location = GenerateRandomSpawnLocation(`SPAWNMGR.GetCurrentXComLocation(), IdealSpawnTilesOffset);
+				Location = GenerateRandomSpawnLocation(GetXComLocation(), IdealSpawnTilesOffset);
 		
 				class'XComGameState_AIReinforcementSpawner'.static.InitiateReinforcements(
 					EncounterID,
@@ -252,8 +252,8 @@ function SpawnReinforcements(int ReinforceIndex, XComGameState NewGameState)
 			// LWOTC: Since the reinforcement spawner doesn't seem to do a good job of
 			// randomisation spawn locations, we manually randomise the spawn location
 			// and use a smaller tile offset for InitiateReinforcements().
-			Location = GenerateRandomSpawnLocation(`SPAWNMGR.GetCurrentXComLocation(), IdealSpawnTilesOffset);
-	
+			Location = GenerateRandomSpawnLocation(GetXComLocation(), IdealSpawnTilesOffset);
+	 
 			// LWOTC: It appears that the second argument to InitiateReinforcements should
 			// be -1, not 0, in order to use the reinforcement countdown of the encounter.
 			class'XComGameState_AIReinforcementSpawner'.static.InitiateReinforcements(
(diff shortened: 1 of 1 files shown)


M3r1st:
Is there any confirmation that the issue you are describing is caused by this specific function call?

Edit: Accoriding to this comment the "position that is out of bounds" assumption you are making might be incorrect:
```
// LWOTC: Pick a random location that is at a distance of TileOffset tiles
// from the given anchor point. This location is guaranteed to be within the
// game board, but it may not necessarily be a valid spawn point.
static function Vector GenerateRandomSpawnLocation(Vector AnchorPoint, int TileOffset)
```

Puppyguarb:
> Is there any confirmation that the issue you are describing is caused by this specific function call?
> 
> Edit: Accoriding to this comment the "position that is out of bounds" assumption you are making might be incorrect:
> 
> ```
> // LWOTC: Pick a random location that is at a distance of TileOffset tiles
> // from the given anchor point. This location is guaranteed to be within the
> // game board, but it may not necessarily be a valid spawn point.
> static function Vector GenerateRandomSpawnLocation(Vector AnchorPoint, int TileOffset)
> ```

If you take `SPAWNMGR.GetCurrentXComLocation()` on that save and check with `WorldData.Volume.EncompassesPoint` you'll get false. I did so much logging to figure that out, I originally thought the issue was with `GenerateRandomSpawnLocation`, but it's with `SPAWNMGR.GetCurrentXComLocation()`.

So just to clear up any confusion, `GenerateRandomSpawnLocation` is working as intended, provided the given anchor point is not out of bounds to begin with.

Tedster59:
Should we instead add some sort of fail-safe to short circuit the infinite loop instead so it can iterate positions? or maybe we need a more robust spawn location system than `WorldData.GetFloorZForPosition`

ToallaNova-O2D:
> Should we instead add some sort of fail-safe to short circuit the infinite loop instead so it can iterate positions? or maybe we need a more robust spawn location system than `WorldData.GetFloorZForPosition`

The issue would be to catch "`SPAWNMGR.getcurrentxcomposition()" and see the return value when out of bounds (to have a reference on how to identify OoB). If the loop is in the native function, it's FUBAR.

Tedster59:
based on the description, the infinite loop is in LW code as mentioned. `GenerateRandomSpawnLocation` likely loops if the Z level returned by `WorldData.GetFloorZForPosition` can never be in the world volume, as it will only check points on a circle of radius R around the out of bounds position, so if this circle never gets anything with an in-bounds Z, it will loop.

M3r1st:
If the issue lies in `GenerateRandomSpawnLocation()`, then that's what needs to be updated; not the logic of getting the initial vector.

Puppyguarb:
> based on the description, the infinite loop is in LW code as mentioned. `GenerateRandomSpawnLocation` likely loops if the Z level returned by `WorldData.GetFloorZForPosition` can never be in the world volume, as it will only check points on a circle of radius R around the out of bounds position, so if this circle never gets anything with an in-bounds Z, it will loop.

Yes, the issue is in the LW code. The x and y can also be out of bounds if the anchor point is far enough out of bounds to begin with, so `WorldData.GetFloorZForPosition` is not the problem. Specifically `GenerateRandomSpawnLocation` never finds an in-bound position because it's too far out of bounds in the first place.

> If the issue lies in `GenerateRandomSpawnLocation()`, then that's what needs to be updated; not the logic of getting the initial vector.

The logic of getting the initial vector is the reason for the issue. The only reason `GenerateRandomSpawnLocation()` starts looping infinitely is because the function is fed a position that is out of bounds.

> Should we instead add some sort of fail-safe to short circuit the infinite loop instead so it can iterate positions? or maybe we need a more robust spawn location system than `WorldData.GetFloorZForPosition`

This is a better solution, if you make it strong enough. I tried making something to slowly expand the circle radius and it didn't work, but I didn't try to code anything too crazy once I figured out the problem was causes by the initial vector.

If you'd like, feel free to close this PR w/o merging and I can make a separate issue to join the other 198. My game is running great now with this fix and my own build!!! No freezes after like a dozen missions, so I'm very happy with this fix but it does not need to join your codebase if you do not want it to.

ToallaNova-O2D:
I think the issue is not that the PR is wrong per se, but rather to find a more generalized fix for all cases (as you said, it can also be found within LWTacticalMissionUnitSpawner, and there might be other circumstances where it presents itself).

furudee:
>Here is the game save that was having this issue, feel free to use it for testing but again there are quite a few mods used in it: [game_save.zip](https://github.com/user-attachments/files/30601607/game_save.zip)

you loaded a save and dlc 0 added old world hero to your HQ barracks and starts rolling bond levels which caused crew to be added to tactical history
and since they are in history they count toward location calculation

Tedster59:
I may just bring this in for now because that would help with this specific issue, since even if I just start adding more and more short circuits or expansion to the existing infinite loop, we'd probably still get strange results, as much as I hate denativizing the function.

PR #1983, opened 2026-08-27
add EffectNames to Shockwave and Shellshock effects

Styrium:


Styrium:
Concussion Warheads could also be moved to a helper function but I didn't touch _that_ bit of code (and am lazy)

PR #1987, opened 2026-09-05
Update markup for item weight

M3r1st:
Move mobility penalty display for item weight to the new DLCInfo hook.

Continuation of #1982


Tedster59:
Will merge this once CHL release hits the main workshop listing.

